PR #1, opened 2017-01-26
Use released nbd-2.1.2 rather than mseri's branch

jonludlam:


jonludlam:
Hmm, there's still a reference to nbd.mseri#ppx in the sources - I'm not sure I understand the workflow.

lindig:
This looks good but needs a change in sources.txt which is where we download file. The reason we need this is such that we can use a different name than the one in the `url` file to keep them unique. I'll take care of this.

I am planning to give this repo some polish and to improve documentation.

jonludlam:
ta

PR #4, opened 2017-02-01
Remove unnecessary OPAM packages from repo

gaborigloi:
The following components were succesfully built without these OPAM
packages:

make forkexecd message-switch gpumon rrdd-plugins sm-cli squeezed v6d \
  wsproxy xapi xcp-networkd xcp-rrdd xenops-cli xenopsd rrd2csv vhd-tool \
  xapi-storage xapi-storage-script

The OPAM package of xapi-storage also depends on the caml2html OPAM
package, but this is a runtime dependency, and we do not make the "html"
target in the specfile, so I did not include "caml2html" in the repo.


gaborigloi:
We will need to update the two generated spec files as well in `xenserver-specs`.

lindig:
Let's try this! I suspect that we might have to add ppx-related packages once we use PPX for generating RPC stubs but it will be good enough to do it when we get there.

PR #7, opened 2017-02-03
Tag backtrace and forkexecd and use stable tagged version

mseri:


lindig:
How about using a shorter version for the directory? 

```
xs/xapi-backtrace.0.3.ppx
```

The way this works is that everything up to the first dot is the package and the rest is the version, as fas as Opam is concerned. But since we make sure to only require the package and only have one version in our universe, this can be kept simple.

mseri:
Makes perfect sense. I am rebuilding it now.

PR #11, opened 2017-02-07
Move packages to the correct upstream/xs directory

gaborigloi:
Ensure that packages are in the correct directory where they belong.

I assumed that all GitHub repos in the xapi-project organisation are maintained by the Xapi project, even if they have many community contributions, and similarly, I assumed that repos outside xapi-project are "upstream", even if they have many contributions from Xapi project members.

I successfully built the following components after this change:
```
make forkexecd message-switch gpumon rrdd-plugins sm-cli squeezed v6d wsproxy xapi xcp-networkd xcp-rrdd xenops-cli xenopsd rrd2csv vhd-tool xapi-storage xapi-storage-script
```


gaborigloi:
I think we have pinned everything now, ran a quick and inaccurate check to confirm that:

```
~/s/xs-opam> cat **/url | grep -v checksum | grep -v .tar.gz | grep -v .tbz
http:
http:
http:
git: "git://github.com/jonludlam/ocaml-rpc#ppx2"
git: "git://github.com/jonludlam/shared-block-ring#ppx"
git: "git://github.com/jonludlam/xcp-idl#ppx"
git: "git://github.com/jonludlam/xen-api-client.git#new-ocaml-rpc"
~/s/xs-opam>
```

And the unpinned packages listed here are also listed on the relevant `CP-20731` ticket.

gaborigloi:
@lindig @jonludlam Two packages I'm unsure about are `xenstore` and `xenstore_transport`: these are in the mirage/ocaml-xenstore and djs55/ocaml-xenstore-clients repos respectively, neither of them is in the xapi-project organization. I did not move them into the `upstream` directory from `xs/` this time, because they are quite xapi-related, and I'm not sure whether these are upstream projects.

PR #12, opened 2017-02-07
Fix LICENSE to name Citrix, add support for mirroring

lindig:
The xs-opam repository was moved from github.com/lindig to
github.com/xapi-project recently but we forgot to update the license at
the time.

The sources.rb script computes the URL from where to download a
package's source code. It now accepts a flag such that it can take
mirroring into account.

Signed-off-by: Christian <christian.lindig@citrix.com>

### DIFF
--- a/utils/sources.rb
+++ b/utils/sources.rb
@@ -84,9 +84,22 @@ opts.each do |opt, arg|
       puts <<-EOF
       sources.rb [option] path/package/url ..
 
-      --url, -u                               just emit URLs
-      --mirror http://example.com/some/path   download from mirror
-      --help                                  show this help
+      -u, --url                                   just emit URLs
+      -m, --mirror http://example.com/some/path   download from mirror
+      -h, --help                                  show this help
+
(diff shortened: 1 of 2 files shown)


utils/sources.rb
```diff
@@ -61,24 +69,30 @@ def to_s
...
 opts.each do |opt, arg|
   case opt
     when '--help'
       puts <<-EOF
       sources.rb [option] path/package/url ..
 
-      --url, -u:    just emit URLs
+      --url, -u                               just emit URLs
+      --mirror http://example.com/some/path   download from mirror
+      --help                                  show this help
```
gaborigloi: This is a really minor thing, but I was wondering whether the shorthand `-m` and `-h` switches for `--mirror` and `--help` are deliberately omitted from here?  (Although I'm not sure how we could list both switches for `--mirror` as it has an argument as well.)

lindig: I'll add them and then merge.

PR #13, opened 2017-02-08
Apply some polish to README and Makefile, support d/l mirror in Makefile

lindig:
These two commits add a bit of polish:

* The invocation of sources.rb is now covered by a macro in the Makefile. This makes it easier to add new name spaces (say `xs-daemons`) and adding a mirror for downloads.

* Downloading from a mirror is now supported by the Makefile and explained in the help message of sources.rb.

* Handling download mirrors in sources.rb is better explained in the help message.


utils/sources.rb
```diff
@@ -84,9 +84,9 @@ def to_s
       puts <<-EOF
       sources.rb [option] path/package/url ..
 
-      -u, --url                                   just emit URLs
-      -m, --mirror http://example.com/some/path   download from mirror
-      -h, --help                                  show this help
+      -u, --url                                     just emit URLs
```
gaborigloi: With the above changes in the Makefile, this `--url` option isn't really needed anymore, if I'm not mistaken? But I'm fine with keeping it though.

Makefile
```diff
@@ -5,30 +5,32 @@
...
 
 #
 # build an Opam repo in build/ with all URL files pointing
 # to sources in build/src.
 #
 repo:	build
 	cp -r packages build
-	./utils/sources.rb packages/*/*/url | while read pkg url; do \
+	$(SRCS) | while read package url; do \
 		echo "http: \"file://$(SRC)/$$(basename $$url)\"" > build/packages/$$pkg/url;\
```
gaborigloi: Indeed, I ran `make repo`, and the URLs did not get rewritten:
```fish
~/s/xs-opam> cat build/packages/url
http: "file:///home/gabori/src/xs-opam/build/src/ocaml-xenstore-clients-0.9.4.tar.gz"
~/s/xs-opam> cat build/packages/xs/xenstore_transport.0.9.4/url
archive: "https://github.com/djs55/ocaml-xenstore-clients/archive/0.9.4.tar.gz"
checksum: "470299fe940fc4b67dfeeab430c06b1d"
```

utils/sources.rb
```diff
@@ -96,16 +96,20 @@ def to_s
...
 
-      and the mirror is http://mirror.example.com/x/y, the resulting URL
+      and the mirror is http://mirror.example.com/x/y/, the resulting URL
       will be
 
         http://mirror.example.com/xy/y/pack.tar.gz
+
+      Note that the mirror must end with a slash or otherwise the last
+      element of the path will be ignored. It is possible to provide the
+      --mirror flag without an argument - in which case no mirrow will be used. In the future a default mirror might be used in that case.
```
gaborigloi: Maybe this line could be broken, as it is quite long.

utils/sources.rb
lindig: I'd like to keep it - I just gave up on using it to make things more uniform.

Makefile
lindig: Good catch! Will fix it.

lindig:
I've addressed the comments. I've also added one commit that renames the package for message-switch because the name did not reflect that we are using a fixed version.

gaborigloi:
@lindig hmm, cannot see that commit here though, will that be part of another PR? (There is a small typo in the last commit message: `break long line in ...`, but it isn't that important.)

PR #15, opened 2017-02-09
Upated libraries xcp-idl and shared-block-ring

lindig:
ce838a1 2017-02-09 xcp-idl: new release 1.90.1 [Christian]
2383641 2017-02-09 shared-block-ring: fix to version 2.3.0 [Christian]


gaborigloi:
Looks good, just a few comments:

* This PR updates xcp-idl from 1.90.0 to 1.90.1, but it seems to me that v1.90.0 is based on ely, while v1.90.1 is based on trunk. But this is fine, because there haven't been any important changes in xcp-idl since ely.

* I think the `opam` file for xcp-idl should be updated to match the one in the repo at `v1.90.1`, they seem to be different.

lindig:
I was fooled by the small increment of the version number and did not check the `opam` file - will change it.

gaborigloi:
@lindig I did a `diff` between the `opam` file in the `xcp-idl` repo at `v1.90.1` and the one in `xs-opam`, and they seem to be different.
We could even have an automated test to ensure that the `opam` files here match the ones in the source repos.

PR #19, opened 2017-02-17
CP-20761: Update libraries to their trunk versions

gaborigloi:
* Point to the new `ppx_trunk` version of `xapi-idl`, which has been rebased on `master`
* Update the `cdrom` and `fd-send-recv` libraries, which had a newer version on the trunk branch of `baggage` than here and on the `ely` branch of baggage

It seems that no other libraries require updating. For most of these libraries, their spec files in `baggage` or `xenserver-specs` point to the same version on the `trunk/ring3` branch as on the `ely` branch. In addition, as I explained on CP-20761, many of the libraries that we had to change to make `ppx-ely` compile don't need rebasing for various reasons.


packages/xs/xapi-idl.1.90.1/opam
```diff
@@ -10,10 +19,11 @@ remove: [
 depends: [ "uri"
            "re"
            "cmdliner"
-           "cohttp" {>= "0.15.2"}
+           "cohttp"
            "xmlm"
-           "rpc" {>= "1.9.4"}
+           "rpc" {>= "1.5.3"}
```
lindig: Do we need this version constraint? The way xs-opam is constructed, there is only one version per component and either this works or it doesn't.

gaborigloi: No, but there are many other `opam` files in `xs-opam` with version constraints as we usually copy the original opam files into the repo.

I think it is a useful safety check to have these version constraints, because this can warn us in case we include the wrong version of a package that is incompatible with another one in the repo. For example, OPAM would complain if we included rpc 1.0.0 in our repo, instead of the newest version. So it might be useful to preserve these version constraints as a sanity check, as they tell us something about which packages our current one is or isn't compatible with.

gaborigloi: Thanks for pointing out the dependency versions though, looking at the diff again I found a problem with the original `opam` file of `xcp-idl` (`cohttp` is duplicated in the `depends` section).

lindig: I agree that we should not touch the original `opam` file if possible.

gaborigloi:
I've modified the `xapi-idl` commit to only update the necessary parts of the `opam` file, now it should be OK, there is no duplicate `cohttp` line, and the correct `rpc` version is used.

PR #20, opened 2017-02-20
rpc (ocaml-rpc): update to 1.9.9

lindig:
Package ocaml-rpc has been updated to version 1.9.9.


gaborigloi:
I reckon we could/should automate all these checks :laughing:

PR #30, opened 2017-03-08
Update master to use PPX-based RPC scheme

lindig:
This commit includes updates of libraries such that they now use the PPX-based RPC scheme.


lindig:
Travis passed.

PR #31, opened 2017-03-08
travis: test XS packages uninstall scripts

mseri:
Add testing for the often forgotten uninstall scripts

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -35,6 +35,7 @@ shared-block-ring
 xapi-backtrace
 xapi-idl
 xapi-inventory
+xapi-netdev
 xapi-rrd
 xapi-stdext
 xen-api-client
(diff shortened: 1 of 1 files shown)


gaborigloi:
Would it be possible/sensible to use https://github.com/ocaml/ocaml-ci-scripts, which also tests uninstallation as far as I know?

gaborigloi:
Another possibility is to use https://github.com/ocaml/ocaml-ci-scripts, but we would have to create a dummy `opam` file in this repo to make it work, as far as I know.

mseri:
If we can use the logic in travis.sh to write the content of PACKAGE= for us, I would be happy to use directly the ocaml-ci script instead of only grabbing some part of them by hand

gaborigloi:
@mseri I've done an experiment with the old `opam-repo-dev`: https://github.com/gaborigloi/opam-repo-dev/commit/78636108fe494fd2e60e07ccb27d2123e1fc3c89
I've moved it completely to the `ocaml-ci` script in that commit, but I had to create a dummy package to make it work.

PR #32, opened 2017-03-08
Move xenstore, xenstore_transport, rpc, and shared-block-ring into upstream

gaborigloi:
Probably none of these packages is currently 'owned' by us.


lindig:
This is fine. Technically all packages are forming one universe and the hierarchies are purely organisational.

PR #36, opened 2017-03-09
Move fast-moving packages & dependents to xs-extra

gaborigloi:
Move message-switch, xapi-idl and xapi-forkexecd and all the packages
depending on them to packages/xs-extra.


mseri:
Can you mention the ticket name in the commit and PR title?

lindig:
Should we update the `.travis.yml`? It does not build the packages in `xs-extra`.

mseri:
Yes, I think we should install&remove also the packages in xs-extra

gaborigloi:
These packages are already listed in the `XS` variable, we just need to update the comment.

PR #38, opened 2017-03-10
Update README

lindig:
Explain the 3 name spaces: upstream, xs, xs-extra.

Signed-off-by: Christian <christian.lindig@citrix.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,33 +13,23 @@ remote Opam repository:
 
 ## Layout of This Repository
 
-Read this to understand how to make changes
+Packages are organised into namespaces for organisational aspects. They
+are equivalent from the perspective of the Opam user.
 
-    +-- compilers                       part of Opam repo (empty)
-    +-- LICENSE
-    +-- packages                        part of Opam repo
-    |   +-- upstream                    all upstream packages
-    |   |   +-- angstrom.0.1.1
-    |   |   |   +-- descr
-    |   |   |   +-- findlib
-    |   |   |   +-- opam
-    |   |   |   \-- url
-    |   |   \-- zed.1.4
-    |   |       +-- descr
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,33 +13,23 @@ remote Opam repository:
...
-    +-- README.md
+Packages are organised into namespaces for organisational aspects. They
+are equivalent from the perspective of the Opam user.
+
+* `upstream`: packages that we don't control with fixed versions.
+* `xs`: packages that we control, with fixed versions.
+* `xs-extra`: packages that we control, following their respective
+  master branch.
+
+    ./packages
```
gaborigloi: Looking at https://github.com/lindig/xs-opam/, this is not recognized as a code block sadly.

README.md
```diff
@@ -13,33 +13,23 @@ remote Opam repository:
...
-    |       |   +-- descr
-    |       |   +-- opam
-    |       |   \-- url
-    |       \-- xenstore_transport.0.9.4
-    |           +-- descr
-    |           +-- findlib
-    |           +-- opam
-    |           \-- url
-    +-- README.md
+Packages are organised into namespaces for organisational aspects. They
```
gaborigloi: Would it be better to move this paragraph with the bullet points to the bottom, below the layout and general description of `packages/`? Maybe that would improve the flow of the text.

README.md
mseri: You are right! I missed it. We need to add 4 more spaces.

README.md
mseri: I don't know. I think it makes sense to have the information straight-away. But I would be happy either way

README.md
lindig: I've fixed the layout problem and added some small touches.

PR #40, opened 2017-03-13
Add vncproxy to xs-extra

gaborigloi:


gaborigloi:
I did not use `archive:` here in the `url` file as it is not necessary, we don't put the stuff in `xs-extra` into the spec file.

mseri:
Already passed a travis build in the forked repo. Merging

PR #43, opened 2017-03-14
Add wsproxy to xs-extra

mseri:
plus update xenstore opam file

### DIFF
--- a/packages/xs-extra/wsproxy.master/opam
+++ b/packages/xs-extra/wsproxy.master/opam
@@ -1,11 +1,11 @@
 opam-version: "1.2"
 name: "wsproxy"
-version: "1.1.0"
-authors: [ "Jon Ludlam" "jonathan.ludlam@citrix.com"
+maintainer: "xen-api@lists.xen.org"
+authors: [ "Jon Ludlam <jonathan.ludlam@citrix.com>"
            "Marcello Seri <marcello.seri@citrix.com>" ]
 license: "LGPL-2.0 with OCaml linking exception"
 homepage: "https://github.com/xapi-project/wsproxy"
-dev-repo: "https://github.com/xapi-project/wsproxy"
+dev-repo: "https://github.com/xapi-project/wsproxy.git"
 bug-reports: "https://github.com/xapi-project/wsproxy/issues"
 build: [
   ["oasis" "setup"]
@@ -20,9 +20,9 @@ remove: [
   ["ocaml" "setup.ml" "-uninstall"]
 ]
 depends: [
-  "lwt" {build}
+  "lwt"
   "ocamlbuild" {build}
   "ocamlfind" {build}
   "oasis" {build}
-  "re" {build}
+  "re"
 ]
--- a/packages/xs-extra/wsproxy.master/opam
+++ b/packages/xs-extra/wsproxy.master/opam
@@ -1,0 +1,28 @@
+opam-version: "1.2"
+name: "wsproxy"
+version: "1.1.0"
+authors: [ "Jon Ludlam" "jonathan.ludlam@citrix.com"
+           "Marcello Seri <marcello.seri@citrix.com>" ]
+license: "LGPL-2.0 with OCaml linking exception"
+homepage: "https://github.com/xapi-project/wsproxy"
+dev-repo: "https://github.com/xapi-project/wsproxy"
+bug-reports: "https://github.com/xapi-project/wsproxy/issues"
+build: [
+  ["oasis" "setup"]
+  ["make"]
+]
+install: [
+    ["oasis" "setup"]
+    ["make" "install"]
+]
+remove: [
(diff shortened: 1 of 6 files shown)


packages/xs-extra/wsproxy.master/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "1.2"
+name: "wsproxy"
+version: "1.1.0"
+authors: [ "Jon Ludlam" "jonathan.ludlam@citrix.com"
+           "Marcello Seri <marcello.seri@citrix.com>" ]
```
gaborigloi: `opam lint` says that `maintainer:` should be defined, as it's mandatory, maybe it could be `maintainer: "xen-api@lists.xen.org"`?

packages/xs-extra/wsproxy.master/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "1.2"
+name: "wsproxy"
+version: "1.1.0"
+authors: [ "Jon Ludlam" "jonathan.ludlam@citrix.com"
```
gaborigloi: Maybe this should be one string, instead of two: `"Jon Ludlam <jonathan.ludlam@citrix.com>"`?

packages/xs-extra/wsproxy.master/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "1.2"
+name: "wsproxy"
+version: "1.1.0"
```
mseri: Ah nice, I didn't know opam lint. I agree we should run it in travis

packages/xs-extra/wsproxy.master/opam
mseri: This came from oasis2opam. I will fix it as well

packages/xs-extra/wsproxy.master/opam
mseri: Sounds good

packages/xs-extra/wsproxy.master/opam
```diff
@@ -0,0 +1,28 @@
...
+authors: [ "Jon Ludlam" "jonathan.ludlam@citrix.com"
+           "Marcello Seri <marcello.seri@citrix.com>" ]
+license: "LGPL-2.0 with OCaml linking exception"
+homepage: "https://github.com/xapi-project/wsproxy"
+dev-repo: "https://github.com/xapi-project/wsproxy"
```
mseri: Why? That can be used to clone the repo.  I am fixing the rest, there is no harm in changing this as well

packages/xs-extra/wsproxy.master/opam
```diff
@@ -0,0 +1,28 @@
...
+]
+install: [
+  ["ocaml" "setup.ml" "-install"]
+]
+remove: [
+  ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
+  ["ocaml" "setup.ml" "-uninstall"]
+]
+depends: [
+  "lwt" {build}
```
mseri: That part was because I used vncproxy as template

packages/xs-extra/wsproxy.master/opam
gaborigloi: Not sure, but `opam lint` doesn't complain when it ends with `.git`, and that is the format that the other `opam` files use.

packages/xs-extra/wsproxy.master/opam
gaborigloi: Oops, my bad, we should fix the `opam` files of `vncproxy` too, I think `build` should be used for things like `oasis`, but not the dependencies that are compiled into it, but I'm not sure though. Based on the above description, I think we do want it to be recompiled when one of the normal dependencies change, so we should probably drop `{build}` for all of these, except `ocamlbuild`, `ocamlfind`, and `oasis`.

packages/xs-extra/wsproxy.master/opam
mseri: We could add something like `find packages -iname opam -print | xargs -n 1 opam lint` to the `travis.sh` file

packages/xs-extra/wsproxy.master/opam
mseri: Fixed both. They are now passing the `opam lint` check

packages/xs-extra/wsproxy.master/opam
mseri: I am testing it here: https://travis-ci.org/mseri/xs-opam/builds/211281238
We need to fix lots of opam files, I guess it's better to do that in another PR

packages/xs-extra/wsproxy.master/opam
gaborigloi: @mseri PR to fix the `opam` file in the `vncproxy` repo: https://github.com/xapi-project/vncproxy/pull/5

PR #44, opened 2017-03-17
CP-20959 obsolete opam-repo-dev by merging it into xs-opam

lindig:
This PR adds basically all packages from opam-repo-dev into this repository such that xapi and xenopsd can be built: see test.sh. However, the solution is not fully general but relies on the CentOS environment that XenServer components are built in.

A lot of packages could benefit from more poiish as they mix compilation and installation or don't uninstall properly. On the other hand, this commit could be a first step towards a more complete Travis setup based on a Docker container.


packages/xs-extra/xapi-rrdd.master/opam
```diff
@@ -0,0 +1,32 @@
...
+  "xapi-xenops"
+  "io-page"
+  "xen-gnt"
+  "xapi-rrd-transport"
+  "oclock"
+  "ounit"
+  "bisect_ppx"
+  "systemd"
+]
+ocaml-version: [>= "4.01.0"]
```
mseri: This should be 4.02.3 givent that we use `bisect_ppx`. Probably we can drop it altogether

mseri:
As a general comment, maybe we should `sed` the maintainers and replace with somebody that is still in Citrix or the xapi mailing list

mseri:
sorry, closed by mistake

packages/xs-extra/xapi-xenopsd.master/opam
```diff
@@ -0,0 +1,38 @@
...
+  "fd-send-recv"
+  "sexplib"
+  "xenstore"
+  "xenstore_transport"
+  "xenctrl"
+  "oclock"
+  "xapi-inventory"
+]
+depopts: "libvirt"
+ocaml-version: [>= "4.01.0"]
```
mseri: I think we can drop it here as well

lindig:
The one fixed version was in `opam-repo-dev` - I did not check why this one was fixed whereas we otherwise are tracking HEAD. Maybe we should track here HEAD as well given that this is not an important library for us.

lindig:
* Updated maintainers
* Removed constraints on OCaml versions in `xs/` and `xs-extra`
* `xapi-storage-ffs` is now tracking master

PR #46, opened 2017-03-21
Add ezxenstore library

jonludlam:


packages/xs/ezxenstore.0.1.0/url
```diff
@@ -0,0 +1,2 @@
+https://github.com/xapi-project/ezxenstore/archive/v0.1.0/ezxenstore-0.1.0.tar.gz
```
gaborigloi: [ERROR] File "/home/opam/.opam/repo/xs-opam/packages/xs/ezxenstore.0.1.0/url", line 1, character 6-7: '/' is not a valid token.
Fatal error:
OpamTypesBase.Lexer_error("'/' is not a valid token")

gaborigloi: I think this should be
```
archive: "https://github.com/xapi-project/ezxenstore/archive/v0.1.0/ezxenstore-0.1.0.tar.gz"
```

jonludlam: ah oops

gaborigloi:
We could also add this package to `travis.sh` to have Travis build it.

lindig:
Maybe we should move to a blacklist for the xs-hierarchy?

lindig:
I'll add the blacklist mechanism in a separate PR - currently testing it.

gaborigloi:
I'm trying to `opam install ezxenstore`, but I keep running into this error:
```
# File "_tags", line 59, characters 7-21:
# Warning: the tag "pkg_lwt.syntax" is not used in any flag or dependency declaration, so it will have no effect; it may be a typo. Otherwise you can use `mark_tag_used` in your myocamlbuild.ml to disable this warning.
# ocamlfind ocamldep -package xenstore.client -package xenstore -package unix -package lwt.unix -package lwt -syntax camlp4o -modules lib/xs_transport.ml > lib/xs_transport.ml.depends
# + ocamlfind ocamldep -package xenstore.client -package xenstore -package unix -package lwt.unix -package lwt -syntax camlp4o -modules lib/xs_transport.ml > lib/xs_transport.ml.depends
# ocamlfind: Using -syntax, but no package is selected specifying a preprocessor as required for -syntax
# Command exited with code 2.
# Makefile:17: recipe for target 'build' faile
```

jonludlam:
This problem was because camlp4 wasn't installed, and so the lwt syntax extension wasn't compiled. xenstore_transport either needs to specify a dependence on camlp4 or we need to remove that dependence!

mseri:
It is actually an easy fix, it's just the old camlp4 lwt syntax. I will raise a PR to djs

gaborigloi:
@mseri : https://github.com/djs55/ocaml-xenstore-clients/pull/7

gaborigloi:
@jonludlam Do we also need an `oasis {build}` dependency in the `opam` file?

jonludlam:
done: https://github.com/xapi-project/ezxenstore/commit/817edfdff76f22dae83d100cd79b2084165b3c55

gaborigloi:
@jonludlam Thanks, I think it would be best to move the `v0.1.0` tag to ensure consistency between the `opam` files.

jonludlam:
No, I don't like moving tags in general, and I don't think this is important enough to warrant it.

jonludlam:
I've released a new version (which also doesn't try to install the test binary)

PR #47, opened 2017-03-21
Sync opam file of message-switch

gaborigloi:
After https://github.com/xapi-project/message-switch/pull/29, update the `opam` file here to match the one in its source repository, at the HEAD of the master branch.

PR #51, opened 2017-03-21
Add constaint to message-switch

jonludlam:

### DIFF
--- a/packages/xs-extra/message-switch.master/opam
+++ b/packages/xs-extra/message-switch.master/opam
@@ -19,7 +19,7 @@ depends: [
   "uri"
   "re"
   "mtime"
-  "mirage-block-unix"
+  "mirage-block-unix" {= "2.4.0"}
   "shared-block-ring"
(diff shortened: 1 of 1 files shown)


mseri:
I think this is already contained in https://github.com/xapi-project/xs-opam/pull/47
do we need =?

jonludlam:
well we probably need a minimum at least. I just got a build error when trying to build against 2.1.0: https://travis-ci.org/jonludlam/xcp-idl#L829

mseri:
I am happy to add `=`, please fix the merge conflict. However the travis build is wrong in general because it tries to use the global opam on top of xs-opam instead of using only xs-opam (using EXTRA_REPO in place of BASE_REPO in the travis file)

jonludlam:
true, but I plan to put some of these opam packages into the main opam-repository, in which case we'll need these sorts of constraints.

jonludlam:
Gabor's fix should work for me

PR #53, opened 2017-03-21
rrd2csv: introduce to xs-extra

mseri:
This is not whitelisted in the tests. It requires xapi dependencies.

### DIFF
--- a/packages/xs-extra/rrd2csv.master/url
+++ b/packages/xs-extra/rrd2csv.master/url
@@ -1,0 +1,1 @@
+git: "https://github.com/xenserver/rrd2csv"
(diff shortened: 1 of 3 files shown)


packages/xs-extra/rrd2csv.master/url
```diff
@@ -0,0 +1 @@
+git: "https://github.com/xenserver/rrd2csv"
```
lindig: Can we use an archive url? 
```
archive: "https://github.com/xenserver/rrd2csv/archive/master/master.tar.gz"
```

Would the `master` branch be correct?

mseri: I think so. Fixing it

PR #55, opened 2017-03-22
upstream-extra: introduce ocveralls for coverage builds

mseri:
The version appearing in the URL is 3.3.4, however this is the URL used
in the official opam repository, the hash is the correct one and the
content is the correct one. The original can be seen here: 
https://github.com/ocaml/opam-repository/tree/master/packages/ocveralls/ocveralls.0.3.4

### DIFF
--- a/packages/upstream-extra/ocveralls.0.3.4/url
+++ b/packages/upstream-extra/ocveralls.0.3.4/url
@@ -1,0 +1,2 @@
+archive: "https://github.com/sagotch/ocveralls/archive/3.3.4.tar.gz"
+checksum: "5ba0aee3781c7c85e56f9525a56a182f"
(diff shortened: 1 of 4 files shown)


packages/upstream-extra/ocveralls.0.3.4/url
```diff
@@ -0,0 +1,2 @@
+archive: "https://github.com/sagotch/ocveralls/archive/3.3.4.tar.gz"
```
lindig: The archive is for 3.3.4 but the package has version 0.3.4.

mseri: Yes, it's the official opam url & version. `0.3.4.tar.gz` has the wrong hash

mseri: Here for the original files: https://github.com/ocaml/opam-repository/tree/master/packages/ocveralls/ocveralls.0.3.4

PR #60, opened 2017-03-27
Fix uninstall of netdev and tapctl

jonludlam:

### DIFF
--- a/packages/xs-extra/xapi-netdev.master/opam
+++ b/packages/xs-extra/xapi-netdev.master/opam
@@ -12,8 +12,6 @@ install: [
   [make "install" "BINDIR=%{bin}%"]
 ]
 remove: [
-  ["oasis" "setup"]
-  [make "uninstall" "BINDIR=%{bin}%"]
   ["ocamlfind" "remove" "netdev"]
 ]
 available: [ os = "linux" ]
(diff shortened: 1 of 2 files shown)


packages/xs-extra/xapi-netdev.master/opam
```diff
@@ -12,8 +12,6 @@ install: [
   [make "install" "BINDIR=%{bin}%"]
 ]
 remove: [
-  ["oasis" "setup"]
-  [make "uninstall" "BINDIR=%{bin}%"]
```
gaborigloi: Oh, so running the `make uninstall` provided by OASIS isn't even necessary, is it sufficient to only run `ocamlfind remove`? https://github.com/xapi-project/xs-opam/pull/34/files and https://github.com/xapi-project/netdev/pull/8 fixed the uninstallation of `netdev` by adding the `oasis setup` line.

jonludlam: This package _only_ provides a library, and to rerun oasis seems unnecessary

lindig: In general, a `make uninstall` is required because a package may install binaries and documentation that is not managed by `ocamlfind` and hence would not removed by it. If a package only installs a library, `ocamlfind` can be enough to remove it.

gaborigloi:
I'll send two PRs to update the `opam` files in the source repos.

gaborigloi:
PRs:
* sync `netdev`: https://github.com/xapi-project/netdev/pull/9
* sync `tapctl`: https://github.com/xapi-project/xs-opam/pull/68

PR #64, opened 2017-03-31
Add ezxenstore dependence to xapi-libs-transitional

jonludlam:


packages/xs-extra/xapi.master/opam
```diff
@@ -43,6 +43,7 @@ depends: [
   "xapi-rrdd-plugin"
   "pci"
   "sha"
+  "xenopsd"
```
lindig: In xs-opam, the package is called `xapi-xenopsd`.

mseri: I am testing @lindig suggestion locally, and indeed it compiles if we add "xapi-xenopsd" as a dependency

mseri: Will we need to add `ulimit -s unlimited` or `OCAMLRUNPARAMS` to avoid segfaulting on the datamodel?

lindig: I'd say we add `OCAMLRUNPARAM='l=8G'` into the environment.

mseri:
All packages are now installing :-) we need to fix uninstalls now

lindig:
Do we have to update the README and remove the script?


>This however will not contain all the packages. To have a full-fledged opam repository you can proceed as follows:

```
git clone https://github.com/xapi-project/xs-opam.git
cd xs-opam
sh into_repo.sh
opam repo add xs-opam .
```

mseri:
No. The packages are still requiring xenctrl, right?

lindig:
I assume. Wasn't sure about the new configure switch for xenopsd that you implemented.

mseri:
Ah, no, that's just to allow the compilation of xenopsd with newer versions of xen (and more recent linux distributions)

PR #72, opened 2017-05-03
Make `xs-opam` a full-fledged opam repo deprecating `into_repo.sh`

mseri:


.dockerignore
```diff
@@ -0,0 +1 @@
+.git/*
```
lindig: I assume `.git/` would be better. The difference is that `.git/.foo` would not be matched by `.git/*` (I believe).

mseri:
That's not going to be a problem. If you only look at `xs` and `upstream`, there is still just one `xenctrl`. When you use the `-extras`, then you have to and `xs-extra/xenctrl.master` is considered "newer" than `xs/xenctrl.dummy`.

PR #75, opened 2017-05-04
Fix breakable travis builds

mseri:


travis.sh
```diff
@@ -34,19 +34,24 @@ else
...
     find packages -iname opam -print | xargs -n 1 opam lint
 else
     opam install -y depext
-    opam depext  -y $UPSTREAM $XS
-    opam install -y -j 4 $UPSTREAM $XS
-    # Workaround to mark failed uninstall as error. We only test
-    # the uninstall of the xs packages but not of the upstream packages.
-    opam remove  -y $XS
-    opam install -y -j 4 $XS
+    if [ "${OCAML_VERSION}" = "4.02" ]; then
```
gaborigloi: Why do we only test uninstallation of packages when ` "${OCAML_VERSION}" = "4.02"`?

gaborigloi:
I assume it's intentional that we did not add `lwt < 3.0.0` to the packages in `xs/` that depend on lwt - do these work with lwt 3.0.0?

travis.sh
mseri: For newer compilers I think it's more important to know if we can actually compile xapi packages. Also, uninstallation should be working (or broken) independently of the compiler version.

mseri:
Concerning `lwt`. I think we should constraint the packages to work with the upstream opam as well. In this way we can have `merlin` and `ocp-*` stuff alongside our packages. We can always create devel or test branches with such constraints when we decide to port/upgrade the packages.

travis.sh
gaborigloi: Oh ok, thanks for the explanation, I just didn't understand why this condition is used.

mseri:
The missing step to make this PR properly useful, is to update the opam files that are failing uninstallation and check if we are still using libvirt anywhere

gaborigloi:
Sounds good :), we can do that in a subsequent PR, I'm working on fixing `sm-cli` at the moment.

travis.sh
mseri: I added a comment in the travis file to clarify it.

mseri:
The additional builds are green!! Merging :-)

PR #76, opened 2017-05-04
Update opam file of xapi-storage-cli

gaborigloi:

### DIFF
--- a/packages/xs-extra/xapi-storage-cli.master/opam
+++ b/packages/xs-extra/xapi-storage-cli.master/opam
@@ -1,13 +1,26 @@
-opam-version: "1"
+opam-version: "1.2"
 maintainer: "xen-api@lists.xen.org"
+authors: [ "xen-api@lists.xen.org" ]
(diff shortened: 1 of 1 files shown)


gaborigloi:
I'm not sure whether I should add the `base-threads` dependency, as `threads` is a dependency in its `_oasis` file (but I'm not sure why it's required).

gaborigloi:
For some reason the installation of this package is failing.

gaborigloi:
Fixed the issue now, it should install without problems in the Travis build.

mseri:
Given that it's part of the standard distribution, I think it's fine as is.

gaborigloi:
@mseri I looked at the libraries listed as dependencies in its `_oasis` file, and listed them here too as OPAM dependencies.

gaborigloi:
A few of them list `base-threads` as an OPAM dependency.

gaborigloi:
I've added the `base-threads` OPAM dependency.

PR #77, opened 2017-05-05
Update lwt constraint on `xs` packages

mseri:
This prevents `xenctrl.dummy` to be chosen in place of `xenctrl.master` when `lwt 3.0.0` is already installed. 
It should resolve an issue arisen from https://github.com/xapi-project/dev-vm/pull/2 and avoid the need to pin lwt by hand.

### DIFF
--- a/packages/xs/nbd.2.1.2/opam
+++ b/packages/xs/nbd.2.1.2/opam
@@ -27,7 +27,7 @@ depends: [
   "ounit" {test}
   "ocamlfind" {build}
   "ppx_tools" {build}
-  "lwt"
+  "lwt" { < "3.0.0" }
   "cstruct" {>= "1.9.0"}
   "cmdliner"
   "sexplib"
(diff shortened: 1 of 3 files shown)


Frezzle:
Did you do any tests? (e.g. with dev-vm using this updated xs-opam and without pinning lwt)

mseri:
Testing now. Wait before merging

gaborigloi:
@mseri In the past I was thinking about a different approach for making this a full-fledged OPAM repo: rather than having both `xenctrl.dummy` and `xenctrl.master` in the repo, and relying on `dummy < master`, we could have the `xenctrl.system` package description *outside* the `packages` directory, so it won't be included in the OPAM repo, and do some extra steps in the spec file to replace `xenctrl.master` with `xenctrl.system`.
Maybe if we get more errors like this, it might be worth considering alternative solutions like the above.

mseri:
The problem is a bit bigger... From the tests it looks like `aspcud` is picking old versions of our libraries with unconstrained lwt, making the install completely inconsistent.

mseri:
The fact that old versions of the libraries with bad constraints are in the main opam repository is not helping

PR #80, opened 2017-05-08
Revert "remove optcomp 1.6" - required by xenopsd

lindig:
This reverts commit 81b2f8265da39d8ac60c6e5f274102456f3f5aca.

The package is required by xenopsd.


mseri:
We opted to fix it in xenopsd instead. It was a spurious dependency in the `_oasis` file.

PR #83, opened 2017-05-17
CA-245333: Added mustache package to the xapi dependencies.

kc284:
As per request at https://github.com/xapi-project/xen-api/pull/3038


gaborigloi:
Let's merge it together with https://github.com/xapi-project/xen-api/pull/3038

PR #88, opened 2017-05-18
Add xen-api-sdk to xs-opam

gaborigloi:
This will help us notice when we break the SDK generation code.

I already have a fork of `xen-api-sdk` with this `opam` file, and a working Travis build: https://travis-ci.org/gaborigloi/xen-api-sdk/builds/233636569

### DIFF
--- a/packages/xs-extra/xen-api-sdk.master/descr
+++ b/packages/xs-extra/xen-api-sdk.master/descr
@@ -1,0 +1,1 @@
+Xen API SDK generation code.
--- a/packages/xs-extra/xen-api-sdk.master/opam
+++ b/packages/xs-extra/xen-api-sdk.master/opam
@@ -1,0 +1,16 @@
+opam-version: "1.2"
+maintainer: "xen-api@lists.xen.org"
+authors: [ "xen-api@lists.xen.org" ]
+homepage: "https://github.com/xapi-project/xen-api-sdk"
+bug-reports: "https://github.com/xapi-project/xen-api-sdk/issues"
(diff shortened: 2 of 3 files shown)


mseri:
Does this need the mustache change for xapi?

kc284:
I don't think it should need it since it would be installed alongside the xapi dependencies.

gaborigloi:
It only needs that change once that PR is merged I think, but I based the `opam` file on the current state of the sdk.

mseri:
Yes, @gaborigloi that's what I was asking. This will be merged only after https://github.com/xapi-project/xs-opam/pull/83

Why is the mustache requirement in xapi's opam file and not in this one?

gaborigloi:
@mseri Oh ok, I thought this would be merged before https://github.com/xapi-project/xs-opam/pull/83, I've prepare this so that it works with the current state of the SDK, as I couldn't base it on that PR

kc284:
This one needs xapi, so the dependency would be installed anyway. But as @gaborigloi correctly pointed out, the dependency should be listed in the sdk anyway in case it's removed in future from xapi.

mseri:
@gaborigloi if we merge it now we break all the additional builds in travis.

kc284:
The PR containing the addition of mustache is not merged yet https://github.com/xapi-project/xen-api-sdk/pull/95 - you can review it if you want :)

gaborigloi:
I've added the mustached dependency so that this `opam` file will be valid after https://github.com/xapi-project/xen-api-sdk/pull/95 is merged.

PR #91, opened 2017-05-30
Set cstruct < 3.0.0 on all our packages

gaborigloi:
To fix the build against the upstream opam.


gaborigloi:
The cstruct [README](https://github.com/mirage/ocaml-cstruct/blob/master/CHANGES.md) for 3.0.0 says "Transitional findlib packages with the old scheme are available" - does this mean that this change is not needed, and we can simply transition while using 3.0.0?

mseri:
Yes, it is just for findlib. He change affects many old packages, some of which are in our dependencies, so we cannot yet move to it. I hope we will be able to throw in much newer dependencies with th jbuilder port, but this will need to be part of the ppx part 2 as it includes many things that switched to result

PR #93, opened 2017-05-31
Update rpc to upstream version 1.9.51

gaborigloi:
Now we can use the upstream version, the changes from our fork have been
merged.


packages/upstream/rpc.1.9.51/url
```diff
@@ -0,0 +1,2 @@
+archive: "https://github.com/jonludlam/ocaml-rpc/archive/v1.9.51.tar.gz"
```
mseri: This should now be `mirage/ocaml-rpc`

gaborigloi: Thanks, fixed

PR #94, opened 2017-06-02
Add xen-gnt < 3.0.0 bound to all our packages

gaborigloi:
Version 3.0.0 has been split into two OPAM and ocamlfind packages.


gaborigloi:
The builds now look better: https://travis-ci.org/gaborigloi/xs-opam/builds/238722711

mseri:
We have a number of random failures, but they are independent of this

PR #95, opened 2017-06-02
Build systems cleanup

mseri:
- Remove omake and obuild that were not used by any package.
- Fix a couple of opam files, one of which for a _pure python_ repository
- Introduce jbuilder


gaborigloi:
These still seem to use `obuild` :( :
```
ackages/xs-extra/xapi-plugin.master/opam
12:  "obuild" {build}
packages/xs-extra/xapi-storage-ffs.master/opam
16:  "obuild" {build}
```

packages/xs-extra/xapi-plugin.master/opam
```diff
@@ -9,6 +9,8 @@ remove: [
   ["ocamlfind" "remove" "worker"]
 ]
 depends: [
-  "obuild" {build}
+  "oasis" {build}
+  "ocamlbuild" {build}
+  "base-unix"
   "rpc"
```
gaborigloi: We could have `"rpc" {>= "1.9.51"}` here, but since it probably won't be upstreamed, it's not important to add this constraint.

euanh:
:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

packages/xs-extra/xapi-storage-ffs.master/opam
```diff
@@ -6,14 +6,4 @@ build: [
...
-  "xapi-idl"
-  "re"
-  "rpc"
-  "ocamlfind"
-  "cmdliner"
-  "cohttp" {< "0.22.0"}
-  "syslog"
-  "obuild" {build}
-  "vhd-format"
-  "xapi-tapctl"
```
gaborigloi: We could consider removing this package, as it's Python-only.

packages/xs-extra/xapi-plugin.master/opam
mseri: I did not do it because we only use the legacy rpclib functions in there, no new ppx-related stuff, so it makes no difference in the end

gaborigloi: Oh you're right, in that case we should leave it.

PR #101, opened 2017-06-08
Sync opam files with main OPAM repo

gaborigloi:


mseri:
We need to fix the version of our patched mtime to unbreak some things. I will make a PR for that

PR #102, opened 2017-06-09
Rename mtime so that upstream bounds are correct also in xs-opam

mseri:
This unbreaks our full build.


gaborigloi:
Now conduit doesn't compile :(, the OCaml 4.04 job used to work for my PR: https://travis-ci.org/gaborigloi/xs-opam

mseri:
Conduit is not compiling for a while. We need to update for the upstream fix for it, it will be done soon. The PR in coduit was approved tonight

PR #105, opened 2017-06-23
update opam after coverage build changes

edwintorok:
See https://github.com/xapi-project/xenopsd/pull/350

Thanks to @gaborigloi for the reminder and help with the `remove` rule.


packages/xs-extra/xapi-xenopsd.master/opam
```diff
@@ -3,16 +3,16 @@ maintainer: "xen-api@lists.xen.org"
...
-	["oasis" "setup"]
 	["./configure"]
 	[make]
 ]
 install: [
- 	["ocaml" "setup.ml" "-install"]
+ 	["./setup.bin" "-install"]
 ]
 remove: [
         ["oasis" "setup"]
```
gaborigloi: I think we should also remove this line. (And in general, we should make sure that the `opam` files here and in the source repo will be exactly the same.)

PR #107, opened 2017-06-27
xapi-libs-transitional: sync opam file with source repo

gaborigloi:
After https://github.com/xapi-project/xen-api-libs-transitional/pull/33


gaborigloi:
Maybe we should automate this back-and-forth syncing :/

PR #120, opened 2017-08-04
add bindings for libdlm

edwintorok:
unix-errno is needed as a dependency because cstubs gives integer errnos instead of Unix.error.

### DIFF
--- a/packages/xs/dlm.0.1.0/opam
+++ b/packages/xs/dlm.0.1.0/opam
@@ -1,0 +1,28 @@
+opam-version: "1.2"
+version: "1.0"
+maintainer: "xen-api@list.xensource.com"
+authors: ["Edwin Török"]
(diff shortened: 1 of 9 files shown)


packages/xs/dlm.0.1.0/opam
```diff
@@ -0,0 +1,28 @@
...
+bug-reports: "https://github.com/xapi-project/ocaml-dlm/issues"
+dev-repo: "https://github.com/xapi-project/ocaml-dlm.git"
+license: "LGPL-2.1 with OCaml linking exception"
+build: [
+  ["jbuilder" "build" "--only" name "--root" "." "-j" jobs "@install"]
+]
+build-test: ["jbuilder" "runtest" "-p" name]
+
+depends: [
+  "jbuilder" {build}
```
mseri: 🎉

PR #132, opened 2017-09-05
Use docker to slightly speed up Travis build

mseri:
This also reorganises the build targets

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ services:
   - docker
 script:
   - echo FROM ocaml/opam:${DISTRO}_ocaml-${OCAML_VERSION} > Dockerfile.test
-  - echo VOLUME /bd >> Dockerfile.test
-  - echo VOLUME /repo >> Dockerfile.test
+  - echo VOLUME /bd    >> Dockerfile.test
+  - echo VOLUME /repo  >> Dockerfile.test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,27 +1,32 @@
...
-    - sudo apt-get update -qq
-    - sudo apt-get install libgcrypt20 liblzma5 libselinux1
-    - wget "http://archive.ubuntu.com/ubuntu/ubuntu/pool/main/s/systemd/libsystemd0_219-7ubuntu3_amd64.deb"
-    - wget "http://archive.ubuntu.com/ubuntu/ubuntu/pool/main/s/systemd/libsystemd-dev_219-7ubuntu3_amd64.deb"
-    - sudo dpkg -i libsystemd0_219-7ubuntu3_amd64.deb libsystemd-dev_219-7ubuntu3_amd64.deb
-script: bash -ex travis.sh
+sudo: false
+services:
+  - docker
+script:
```
lindig: I think the echo commands should be quoted to give more assurance that we understand what goes into the final file.

lindig:
There are some hard-coded version numbers in the code. I think these are fine for now but it would be great to get rid of  them or to introduce a variable.

mseri:
I think the next step is to fully rewrite the script to address https://github.com/xapi-project/xs-opam/issues/133

PR #134, opened 2017-09-05
nbd: update to 2.3.0

gaborigloi:
This version adds TLS support to the server.


packages/xs/nbd.2.3.0/opam
```diff
@@ -11,18 +11,17 @@ build-test: ["jbuilder" "runtest" "-p" name]
 depends: [
   "jbuilder" {build}
   "ounit" {test}
-  "ppx_tools" {build}
   "lwt" {>= "2.6.0" & < "3.0.0"}
-  "cstruct" {>= "1.9.0" & < "3.0.0"}
+  "ssl"
+  "cstruct" {>= "2.4.0" & < "3.0.0"}
```
mseri: Changing this to `>= "2.3.0"` should be enough to fix the build

gaborigloi: Testing with this Travis build: https://travis-ci.org/gaborigloi/nbd/builds/272124111

PR #135, opened 2017-09-06
cohttp: add conduit < v1.0.0 upper bound

gaborigloi:
To fix the build when using xs-opam as an extra remote.
In conduit v1.0.0 the package has been split into multiple packages.


gaborigloi:
See https://github.com/mirage/ocaml-conduit/blob/master/CHANGES.md

gaborigloi:
The build error is unrelated but has to be fixed - we broke `vhd-tool` when we added TLS support to `nbd`.

mseri:
This has to be fixed because it prevents us to release a new xs-opam, but it should be done in a different PR

PR #136, opened 2017-09-12
Ported vncproxy to build with jbuilder.

kc284:
Signed-off-by: Konstantina Chremmou <konstantina.chremmou@citrix.com>


kc284:
This depends on https://github.com/xapi-project/vncproxy/pull/6

PR #138, opened 2017-09-13
Temporarily revert "nbd: update to 2.3.0"

gaborigloi:
Reverts xapi-project/xs-opam#134

We will reintroduce the nbd version with TLS support once we fix the compilation issue in `vhd-tool`.


thomassa:
Oh, I thought we had done the vhd-tool fix already. Was that only on some other branch?

gaborigloi:
We've only done it for feature/CBT, it was just a quick temporary fix, probably not a proper one.

PR #147, opened 2017-10-03
drop oasis test-only dependencies

edwintorok:
We do not run the tests for oasis, and it would need several packages to be
added to xs-opam to make BASE_REMOTE Travis testing work
(which tries to install the test-only dependencies too).


gaborigloi:
I think we could also drop the `opam` files for all the `base-...` packages, because they should already be installed?

mseri:
I think it's safe to keep the base libraries, they are being moved around in recent compilers

PR #150, opened 2017-10-05
qmp 0.12.0 - new release

lindig:
This release enlarges the type event_data, which is likely to require
clients to handle it.

CP-24774: Add unit test for new XEN_PLATFORM_PV_DRIVER_INFO QMP event message
CP-24774: Remove unused Event._XEN_PLATFORM_PV_DRIVER_INFO event name constant
CP-24774: Parse XEN_PLATFORM_PV_DRIVER_INFO QMP event message


lindig:
This will break xenopsd because it needs to currently doesn't handle the enlarged type.

PR #152, opened 2017-10-09
xen-api-client: release 0.11.0

mseri:
And release now separate libraries `xen-api-client-{async,lwt}` `1.0.0`


mseri:
This may break some dependencies, so let's wait for Travis first and fix them before merging

mseri:
@gaborigloi you will need to update `xapi-nbd` to require `xen-api-client-lwt`
@jonludlam you will need to update the dependencies as well

mseri:
Note: I have already manually tested that it does not break `xcp-networkd` and `xapi`

PR #153, opened 2017-10-10
xapi: introduce new packages after xapi port to jbuild

mseri:
Depends on https://github.com/xapi-project/xen-api/pull/3274 and https://github.com/xapi-project/vncproxy/pull/12

Also updates:
- vncproxy, to use the new xen-api-client
- xcp-idl, recently ported to jbuilder

### DIFF
--- a/packages/xs-extra/xapi-cli-protocol.master/descr
+++ b/packages/xs-extra/xapi-cli-protocol.master/descr
@@ -1,0 +1,4 @@
+The xapi toolstack daemon which implements the XenAPI
+
+This daemon exposes the XenAPI and is used by clients such as 'xe'
+and 'XenCenter' to manage clusters of Xen-enabled hosts.
--- a/packages/xs-extra/xapi-client.master/opam
+++ b/packages/xs-extra/xapi-client.master/opam
@@ -1,0 +1,15 @@
+opam-version: "1.2"
+maintainer: "xen-api@lists.xen.org"
+authors: [ "xen-api@lists.xen.org" ]
+homepage: "https://github.com/xapi-project/xen-api"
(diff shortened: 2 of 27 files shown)


packages/xs-extra/xapi-cli-protocol.master/descr
```diff
@@ -0,0 +1,4 @@
+The xapi toolstack daemon which implements the XenAPI
```
lindig: Minor point but I think `that` instead of `which` would be better. `That` would introduce an essential restriction whereas `which` adds an (optional) qualification.

packages/xs-extra/xapi-client.master/opam
```diff
@@ -0,0 +1,14 @@
...
+bug-reports: "https://github.com/xapi-project/xen-api/issues"
+dev-repo: "https://github.com/xapi-project/xen-api.git"
+
+build: [[ "jbuilder" "build" "-p" name ]]
+
+depends: [
+  "jbuilder" {build & >= "v1.0+beta11"}
+  "xapi-datamodel"
+  "xapi-types"
+]
```
lindig: Missing `\n`

packages/xs-extra/xapi-cli-protocol.master/descr
mseri: The descriptions are all wrong (but xapi's) actually... We should probably discuss what to put in all of them

mseri:
This PR is now ready. The change will be completely effective only once the two mentioned PRs are merged, but it already allows to use the jbuilder version of xapi subpackages locally without requiring xapi

lindig:
This looks good. All essential packages are compiling. There is only `xapi-nbd` not compiling right now and this has nothing to do with the changes.

PR #154, opened 2017-10-10
[CBT merge] Update nbd to 2.5.0

gaborigloi:
And sync the opam file with the one in the source repo.

Changes since version 2.2.0:
* v2.5.0: Support for serving block devices as read-only
    * NBD server: add support for serving block devices as read-only
* v2.4.0: add logging using Lwt_log
* v2.3.2: less restrictive ppx_sexp_conv bounds in opam file
    * nbd.opam: make ppx_sexp_conv bound less restrictive
* v2.3.1: lower cstruct bound in opam file
    * nbd.opam: require cstruct 2.3.0 instead of 2.4.0
* v2.3.0: TLS support for the server
    * TLS support (libs and server only)
    * New functions for ease of safe use


gaborigloi:
Merge together with https://github.com/xapi-project/vhd-tool/pull/45

gaborigloi:
We discussed with Rob that this should be merged on Thursday after the next merge.

PR #155, opened 2017-10-12
xen-api-client: update to v0.12.0

gaborigloi:


packages/xs/xen-api-client.0.12.0/url
```diff
@@ -1,2 +1,2 @@
-archive: "https://github.com/xapi-project/xen-api-client/archive/v0.11.0.tar.gz"
+archive: "https://github.com/xapi-project/xen-api-client/archive/v0.12.0.tar.gz"
 checksum: "3820f857f37ae954e4edadb6f4d40c94"
```
mseri: This needs update

packages/xs/xen-api-client-async.1.0.0/url
```diff
@@ -1,2 +1,2 @@
-archive: "https://github.com/xapi-project/xen-api-client/archive/v0.11.0.tar.gz"
+archive: "https://github.com/xapi-project/xen-api-client/archive/v0.12.0.tar.gz"
 checksum: "3820f857f37ae954e4edadb6f4d40c94"
```
mseri: This is not necessary (similarly for the `-lwt`), but there is no need to revert the change

PR #156, opened 2017-10-17
xen-api-sdk: depend on xapi-datamodel instead of xapi

gaborigloi:


gaborigloi:
This syncs the opam file with the source repo after https://github.com/xapi-project/xen-api-sdk/pull/120

PR #158, opened 2017-10-20
CA-269137: add x509 and dependencies

thomassa:
For reading the subjects from x509 TLS certificates, to return
in the subject fields of the Vdi.get_nbd_info return value.

### DIFF
--- a/packages/upstream/conf-gmp.1/opam
+++ b/packages/upstream/conf-gmp.1/opam
@@ -1,0 +1,22 @@
+opam-version: "1.2"
+maintainer: "nbraud"
+author: "nbraud"
+homepage: "http://gmplib.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "https://github.com/ocaml/opam-repository.git"
+license: "GPL"
+build: [
+  ["sh" "-exc" "cc -c $CFLAGS -I/usr/local/include test.c"] {os != "darwin"}
+  ["sh" "-exc" "cc -c $CFLAGS -I/opt/local/include -I/usr/local/include test.c"] {os = "darwin"}
+]
+depexts: [
+  [["debian"] ["libgmp-dev"]]
+  [["ubuntu"] ["libgmp-dev"]]
+  [["osx" "homebrew"] ["gmp"]]
+  [["centos"] ["gmp" "gmp-devel"]]
+  [["fedora"] ["gmp" "gmp-devel"]]
+  [["openbsd"] ["gmp"]]
+  [["freebsd"] ["gmp"]]
+  [["alpine"] ["gmp-dev"]]
+  [["opensuse"] ["gmp-devel"]]
+]
--- a/packages/upstream/conf-perl.1/opam
+++ b/packages/upstream/conf-perl.1/opam
@@ -1,0 +1,15 @@
+opam-version: "1.2"
+maintainer: "tim@gfxmonk.net"
+homepage: "https://www.perl.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "https://github.com/ocaml/opam-repository.git"
+license: "GPL-1+"
+build: [["perl" "--version"]]
+depexts: [
+  [["debian"] ["perl"]]
+  [["ubuntu"] ["perl"]]
+  [["alpine"]["perl"]]
+  [["nixpkgs"] ["perl"]]
+  [["archlinux"] ["perl"]]
+  [["fedora"] ["perl-Pod-Html"]]
+]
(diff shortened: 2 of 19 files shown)


thomassa:
Once this is in, https://github.com/xapi-project/xen-api/pull/3302 will build successfully so long as it is using the master branch of xs-opam.

packages/upstream/nocrypto.0.5.3/opam
```diff
@@ -0,0 +1,47 @@
+opam-version: "1.2"
```
gaborigloi: If we wanted to use `nocrypto` `0.5.4`, we'd have to upgrade to `cstruct` `2.4.0`. I think this is fine, I'll upgrade this lib anyway in my xs-opam update, which contains the newest cstruct.

lindig:
I see no problem introducing new, high-quality packages as long as existing packages don't change behaviour but would like to see @mseri's opinion. Every package that we add we have to publish and watch for security problems - so there is still a cost. The list here looks reasonable. I was initially worried about Perl but it simply demands Perl to be present - which is almost a given on systems today.

packages/upstream/conf-gmp.1/opam
```diff
@@ -0,0 +1,22 @@
...
+license: "GPL"
+build: [
+  ["sh" "-exc" "cc -c $CFLAGS -I/usr/local/include test.c"] {os != "darwin"}
+  ["sh" "-exc" "cc -c $CFLAGS -I/opt/local/include -I/usr/local/include test.c"] {os = "darwin"}
+]
+depexts: [
+  [["debian"] ["libgmp-dev"]]
+  [["ubuntu"] ["libgmp-dev"]]
+  [["osx" "homebrew"] ["gmp"]]
+  [["centos"] ["gmp" "gmp-devel"]]
```
mseri: Do we have GMP installed on xenserver hosts?

packages/upstream/conf-perl.1/opam
```diff
@@ -0,0 +1,15 @@
+opam-version: "1.2"
+maintainer: "tim@gfxmonk.net"
+homepage: "https://www.perl.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "https://github.com/ocaml/opam-repository.git"
+license: "GPL-1+"
+build: [["perl" "--version"]]
+depexts: [
+  [["debian"] ["perl"]]
```
mseri: I trust that this is installed on xenserver, but I'd like to see a depexts stanza for `centos` as well before merging

packages/upstream/nocrypto.0.5.3/opam
mseri: We have bad experiences with upgrades and downgrades, I'd rather stay with this version and bundle the updates in the big PR you are working on for after the next xenserver release

lindig:
We might have to add RPM dependencies in the SPEC files we generate from this.

gaborigloi:
The `zarith` package requires these `conf-` packages.
It says:
> Implements arithmetic and logical operations over arbitrary-precision integers
The Zarith library implements arithmetic and logical operations over
arbitrary-precision integers. It uses GMP to efficiently implement
arithmetic over big integers. Small integers are represented as Caml
unboxed integers, for speed and space economy.

It seems that perl is only needed for building the package:
```
depends: [
  "ocamlfind"
  "conf-gmp"
  "conf-perl" {build}
]
```
Therefore I think we'll need to have both `gmp-devel` and `perl` as `BuildRequires` in `xs-opam-repo.spec`. We already have `gmp-devel` and `gmp` (?) there as `BuildRequires`, so we only need to add perl.

mseri:
Do we need `gmp` as a `Requires` though?

gaborigloi:
We need to figure out whether `gmp` is statically or dynamically linked into `zarith`. We can try using `ldd` to see if it's dynamically linked.

packages/upstream/conf-perl.1/opam
gaborigloi: Done.

packages/upstream/conf-gmp.1/opam
gaborigloi: It seems that `gmp` is installed on a xenserver:
```
# rpm -q gmp
gmp-6.0.0-12.el7_1.x86_64
```

mseri:
Let's fix the Requires and BuildRequires then

packages/upstream/nocrypto.0.5.3/opam
thomassa: Yes, I agree: the 0.5.4 nocrypto doesn't seem to have anything we need, and upgrading cstruct would meaning having to recompile lots of other things. (That's why I chose nocrypto 0.5.3 for this pull-request.)

gaborigloi: I've upgraded it in my new `xs-opam` anyway :smile: - will push it into a `next` branch of the main `xs-opam`.

thomassa:
Note that we're only using a small part of the `x509` package's functionality, so a lot of the requirements probably aren't needed at runtime: it might be that the `xapi` binary doesn't call the `gmp` system library, and it might even not have anything from `nocrypto` or `zarith` in it, but I don't think it's worth the trouble of checking.

Issue #117, opened 2017-08-03
Port to jbuilder


mseri:
The following packages need porting to `jbuilder`. Please keep this list updated with the packages that you port and release.

Note: if you port packages using `xapi-stdext`, please modify the dependencies and use the new sub-packages instead.

From `packages/xs`:

- [x] cdrom.0.9.3
- [x] crc.2.0.0
- [x] ezxenstore.0.1.2
- [x] fd-send-recv.1.0.5
- [x] nbd.2.2.0
- [x] netlink.0.3.1
- [x] opasswd.1.0.2
- [x] qmp.0.10.0
- [x] xapi-backtrace.0.5
- [X] xapi-inventory.1.0.2
- [X] xapi-rrd.1.1.0
- [X] xapi-stdext.3.0.0
- [x] xapi-test-utils.1.1.0
- [x] xen-api-client.0.11.0

From `packages/xs-extra`:

- [x] message-switch.master
- [X] rrd2csv.master
- [x] rrddump.master
- [x] vhd-tool.master
- [x] vncproxy.master
- [X] wsproxy.master
- [x] xapi-forkexecd.master
- [x] xapi-idl.master
- [x] xapi-libs-transitional.master
- [x] xapi.master
- [X] xapi-nbd.master
- [x] xapi-netdev.master
- [x] xapi-networkd.master
- [x] xapi-plugin.master
- [x] xapi-rrdd.master
- [x] xapi-rrdd-plugin.master
- [x] xapi-rrd-transport.master
- [x] xapi-squeezed.master
- [x] xapi-storage-cli.master
- [x] xapi-storage.master
- [x] xapi-storage-script.master
- [x] xapi-tapctl.master
- [X] xapi-xenopsd.master
- [x] xapi-xenops.master
- [X] xen-api-sdk.master
- [ ] xenctrl.master (repo name ocaml-xen-lowlevel-libs, broken on recent xen -- needs planning and is not really necessary atm) 🍎 

Not in `xs-opam` (due to issues with blktap-devel):

- [x] rrdd-plugins.master (on xenserver/rrdd-plugins)

🍏 : simple port
🍎 : good luck with that

YarsinCitrix:
Can someone check help to add me in?

mseri:
We don't plan to port `xenctrl.master`. All the rest is ported. Thanks to all the people involved @kc284 @YarsinCitrix @gaborigloi @robhoes @euanh @lindig @edwintorok 

[...]

PR #159, opened 2017-10-23
Re-enable install/uninstall checks

gaborigloi:
The OCAML_VERSION environment variable needs to be passed to the container for travis.sh.


gaborigloi:
To be honest we should actually remove these install/uninstall tests anyway when #117 is completed 😄

mseri:
Let's add an issue linking to #117 to drop those checks after we have jbuildered everything

gaborigloi:
Thanks, added #160

PR #170, opened 2017-12-04
Add script for building the repo locally in a container

gaborigloi:
This script allows one to run our Travis build locally.

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,0 +1,15 @@
+# Script for running the tests in a container and producing a Docker image with
+# the "local-build" tag that contains the installed packages.
+
+set -ex
+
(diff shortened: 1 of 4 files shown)


build-docker.sh
```diff
@@ -0,0 +1,15 @@
+# Script for running the tests in a container and producing a Docker image with
+# the "local-build" tag that contains the installed packages.
+
+set -ex
+
```
lindig: This could be simplified:

```
cat <<EOF > Dockerfile.txt
FROM ocaml/opam:${DISTRO}_ocaml-${OCAML_VERSION}
COPY . /xs-opam
...
EOF
```

gaborigloi:
Hmm, the build without the here document changed passed and with that change it failed - maybe I used incorrect syntax? :thinking:

gaborigloi:
Oops, I made a type, renamed the temporary docker file by accident :flushed:

PR #174, opened 2017-12-06
rrd2csv: fix dependency (xapi -> xapi-client)

mseri:


mseri:
There is already a pending PR

PR #177, opened 2017-12-07
pci: remove library as it is currently vendored in xapi

mseri:


gaborigloi:
xapi's `opam` file should be updated to remove the `pci` dependency

PR #181, opened 2017-12-11
netlink: update depexts for Debian 9 Stretch

gaborigloi:
Different packages are required on the new Debian stable.


gaborigloi:
This is a port of https://github.com/xapi-project/ocaml-netlink/pull/7 to xs-opam. This might fix our debian-unstable build.
The opam files are still not in sync - the source one is missing the centos depexts.

PR #182, opened 2017-12-12
CP-26205: Update xs-opam to align with libraries updates

YarsinCitrix:
Signed-off-by: Yarsin He <yarsin.he@citrix.com>


mseri:
Move `vhd-format.master` to `vhd-format.0.9.1` and update the URL to point at `https://github.com/mirage/ocaml-vhd/archive/v0.9.1.tar.gz`
Is `vhd-format` enough or we need to add also `vhd-format-lwt`? From the patch you added I believe they both need to be added with version `0.9.1` and their URL need to be updated

packages/upstream/vhd-format.0.9.1/opam
```diff
@@ -8,7 +8,7 @@ authors: [ "Dave Scott" "Jon Ludlam" ]
 homepage: "https://github.com/mirage/ocaml-vhd"
 bug-reports: "https://github.com/mirage/ocaml-vhd/issues"
 dev-repo: "git://github.com/mirage/ocaml-vhd"
-build: make
+build: [[make "release"]]
 remove: [[make "uninstall"]]
 depends: [
   "ocamlfind"
```
gaborigloi: We need to update this `opam` file: since we're pointing to `vhd-format` 0.9.1, we should use the opam file corresponding to that version: https://github.com/mirage/ocaml-vhd/blob/master/vhd-format.opam

packages/xs-extra/vhd-tool.master/opam
```diff
@@ -25,12 +25,13 @@ depends: [
   "tar-unix"
   "uri"
   "vhd-format" {>= "0.7.0"}
+  "vhd-format-lwt" {>= "0.7.0"}
```
gaborigloi: `vhd-format-lwt` has been introduced starting at version 0.9.0, so we don't need this version bound, because it will always be true.

packages/upstream/vhd-format.0.9.1/opam
mseri: Well spotted!

packages/upstream/vhd-format.0.9.1/opam
```diff
@@ -8,22 +8,15 @@ authors: [ "Dave Scott" "Jon Ludlam" ]
...
   "cstruct" {>= "1.9"}
-  "mirage-types-lwt" {>= "3.0.0"}
-  "ipaddr"
   "io-page"
+  "rresult"
   "uuidm"
-  "ounit"
-  "ppx_tools" {build}
-  "ocamlbuild" {build}
+  "ppx_cstruct" {build}
```
gaborigloi: I've realized that I've made a mistake in the original opam file, I should have added 
```
jbuilder {build}
```
😳

mseri: This needs to be added here as well

gaborigloi:
It needs to be added to `vhd-format` too

gaborigloi:
PR for upstream source repo: https://github.com/mirage/ocaml-vhd/pull/49 once accepted, I'll open one for the main opam repo.

PR #186, opened 2017-12-13
Correct nbd and nbd-lwt-unix version and url

YarsinCitrix:
Signed-off-by: Yarsin He <yarsin.he@citrix.com>


packages/xs/nbd-lwt-unix.4.0.0b1/url
```diff
@@ -0,0 +1 @@
+archive: "https://github.com/xapi-project/nbd/archive/v4.0.0+beta1.tar.gz"
```
mseri: Add the checksum

packages/xs/nbd.4.0.0b1/url
```diff
@@ -0,0 +1,2 @@
+archive: "https://github.com/xapi-project/nbd/archive/v4.0.0+beta1.tar.gz"
```
gaborigloi: The directory name should be exactly `package-name.package-version`, which is `4.0.0+beta1`, as described here, in the "repository format" section: http://opam.ocaml.org/doc/Manual.html#Layout

PR #194, opened 2017-12-20
rebase master and ocaml 4.05 build issue resolve

YarsinCitrix:
Updates for ocaml 4.05 build issues:
- added camlp4.4.05+2
- added ppx_tools.5.0+4.05
- renamed ppx_tools.5.0 to ppx_tools.5.0+4.04

make 4.05 build success mandatory in travis.
Rebased from master.


mseri:
The three added packages will break the main Jenkins build.
Can you please move them to `packages/upstream-4.05`, and modify the travis script to add the `upstream-4.05` to the packages when the compiler is `ocaml-4.05`?

mseri:
Actually you don't need these three packages. As @gaborigloi pointed out, we are building uit with EXTRA_REMOTES, so opam will get those packages from the upstream opam repository

mseri:
To add a bitmore context, the packages in `xs-opam` are a snapshot of the packages in `opam-repository` that has to satisfy three requirements:

- there is _only one_ version for each package
- all the packages in `upstream`, `upstream-extra`, `xs` must track a specific version
- all the packages have to build with the compiler used in production and have to be compatible between each other

Requiring the travis build to work with 4.05 means only that if you add `xs-opam` as an additional repository, `opam` is able to find upstream packages that satisfy the necessary bounds and are compatible with our packages. But `xs-opam` itself, as a unique base remote, has only to provide what is needed for us to compile with `4.04.2` in production withouth having to manually filter manually packages because we have added multiple versions.

mseri:
Also, there is a known bug on ppx_type_conv v0.9.0 that makes it incompatible with ppx_deriving. So it has to be upgraded to v0.9.1 or even v0.10.0

mseri:
Feel free to rebase this on top of `master` and force push it onto the `next` branch

mseri:
A working configuration, rebased on the current master, has been force pushed on the branch. I am closing this PR

PR #195, opened 2017-12-21
forkexec, xapi-forkexecd: update opam files after port to jbuilder

mseri:

### DIFF
--- a/packages/xs-extra/forkexec.master/url
+++ b/packages/xs-extra/forkexec.master/url
@@ -1,0 +1,1 @@
+archive: "https://github.com/xapi-project/forkexecd/archive/master.tar.gz"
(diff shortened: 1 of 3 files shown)


packages/xs-extra/forkexec.master/url
```diff
@@ -0,0 +1 @@
+archive: "https://github.com/xapi-project/forkexecd/archive/master.tar.gz"
```
kc284: Checksum?

mseri: Not for the ones tracking master, they are not versioned, it would change at each merge

mseri:
The failure should be fixed by: https://github.com/xapi-project/xenopsd/pull/435

PR #198, opened 2018-01-09
Update of `xs-extra` after recent library changes

mseri:
Includes:

- xapi-types and xapi-datamodel cleanup
- rrd2csv dependency fix
- xapi-libs-transitional port to jbuilder


lindig:
This looks good but has a lot of details. Let's wait for Travis.

PR #204, opened 2018-01-16
Ported build from oasis to jbuilder for message-switch and xapi-storage-script; introduced jbuildered rrdd-plugins.

kc284:


kc284:
Please do not merge before the spec file PR is ready.

packages/xs-extra/rrdd-plugins.master/opam
```diff
@@ -0,0 +1,26 @@
...
+license: "LGPL-2.1 with OCaml linking exception"
+homepage: "https://github.com/xenserver/rrdd-plugins"
+bug-reports: "https://github.com/xenserver/rrdd-plugins/issues"
+dev-repo: "git://github.com/xenserver/rrdd-plugins"
+build: [[ "jbuilder" "build" "-p" name "-j" jobs ]]
+depends: [
+  "jbuilder" {build}
+  "rrdd-plugin"
+  "xapi-libs-transitional"
+  "unix"
```
mseri: This should not be here

PR #205, opened 2018-01-17
rpc: update to latest upstream version

mseri:


packages/upstream/rpc.2.3.0/opam
```diff
@@ -27,6 +26,7 @@ depends: [
   "oasis" {build}
   "cppo"  {build}
   "async" {< "v0.10.0"}
+  "camlp4"
```
gaborigloi: Is it deliberate that these two differ and that we've added camlp4?

mseri: No, it's a mistake upstream. I'll fix it in the next release

mseri:
Note that this patch is not needed upstream at the moment, and the missing cmx warning will disappear soon after the next release, once the jbuilder port is merged.

PR #207, opened 2018-01-18
Synchronised the xapi opam file with the one in the xen-api repo.

kc284:


lindig:
I am not sure about the depext changes.  However, we should use the same Opam description in the source repo and in the opam repo. And this PR makes them the same.

kc284:
Hmmm, it may be that these deps need to be copied into the source repo. I'm not sure which are the correct ones, and even more so wheb this commit https://github.com/xapi-project/xs-opam/commit/37a18d037e3dafa5c52c5c23cc3163f896677cfe added them while the commit's message talks about removal...

kc284:
I checked with @mseri and it turns out it's the source repo where the deps need to be corrected.

mseri:
I don't understand why travis is failing. I am investigating it on the side. @gaborigloi have you ever seen that kind of failure?

gaborigloi:
No, I haven't 🤔

mseri:
It's failing to install `blktap-devel` :-(

PR #211, opened 2018-01-26
CP-26711: Ported xapi-storage to jbuilder; added library caml2html.

kc284:
https://github.com/xapi-project/xapi-storage/pull/69

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
+# Opam Repository for XenServer
 
 [![Build Status](https://travis-ci.org/xapi-project/xs-opam.svg?branch=master)](https://travis-ci.org/xapi-project/xs-opam)
-
-# Opam Repository for XenServer
 
 This [Opam] repository supports building components in the XenServer
 toolstack implemented in [OCaml].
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -136,7 +135,8 @@ image.
 For example, the [`ezxenstore`](https://github.com/xapi-project/ezxenstore)
 project can be built in the following way using the `xenserver/xs-opam:latest`
 image:
-```
+
+```bash
```
edwintorok: Have you used a particular tool to find these problems, or was it by visual inspection?

kc284: I develop on VSCode with this plugin https://github.com/DavidAnson/vscode-markdownlint.

packages/xs-extra/xapi-storage.master/opam
```diff
@@ -1,25 +1,21 @@
...
-]
+build: [[ "jbuilder" "build" "-p" name "-j" jobs ]]
+build-test: [[ "jbuilder" "runtest" "-p" name "-j" jobs ]]
 depends: [
+  "jbuilder" {build}
+  "ounit" {test}
+  "cow"
   "rpc"
-  "re"
+  "caml2html"
```
edwintorok: Can this be made an optional dependency? We can still build the whole package without it, its just generating the docs that requires it.

packages/xs-extra/xapi-storage.master/opam
```diff
@@ -1,25 +1,23 @@
...
   "rpc"
-  "re"
+  "xmlm"
   "cmdliner"
 ]
+depopts: [
+  "caml2html"
+]
 depexts: [
-  [ ["debian"] ["python"] ]
```
edwintorok: was this dropped because it was not tested, or is there something broken on debian about it?

kc284: I changed it to centos because this is what we're mostly interested in since xenserver is based on it. I also remember having a discussion on this with @mseri a few days ago.

edwintorok: Well I have python installed anyway on my Ubuntu dev machine, so in this case probably doesn't matter, for other packages we should probably try to keep the debian/ubuntu entries working since that is what our development environment is based on (`dev-vm` etc).

kc284: I think I also need to add `camlmix` to the repo because `caml2html` depends on it.

kc284: Ok I restored the original distros.

PR #214, opened 2018-01-29
fix debian test

mseri:

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -17,8 +17,11 @@ pkg()
     popd > /dev/null
 }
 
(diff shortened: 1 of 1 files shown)


travis.sh
```diff
@@ -17,9 +17,12 @@ pkg()
     popd > /dev/null
 }
 
-if [ "${DISTRO}" = "centos-7" ]; then
```
lindig:
```
case "${DISTRO}" in
centos*)
    ;;
debian)
    sudo apt-get update -y
    ;;
*)
   echo "unknown DISTRO=${DISTRO}"
   exit 1
esac
```

Issue #133, opened 2017-09-05
Automate the Travis build checks


mseri:
As suggested by @gaborigloi, the new docker based Travis build allow for a greater flexibility and could be useful for a number of additional automated checks:

- [x] build a container with all xs-opam installed and push it to the docker repository -> done: https://github.com/xapi-project/xs-opam/pull/165
- [ ] build a container without the `upstream` folder, install `xs` and `xs-extra` and check that we have only needed packages in `upstream` and `upstream-extra` -> detecting unused `upstream` packages done: https://github.com/xapi-project/xs-opam/pull/249 , unused `upstream-extra` packages still not detected
- [x] build a container with `xs-opam` as the extra remote and check which packages we can update, presenting the list as the final stage of the Travis build -> https://github.com/xapi-project/xs-opam/pull/223
- [ ] build and push the complete documentation of all the packages we use
- [x] make it possible to build the container locally -> done: https://github.com/xapi-project/xs-opam/pull/170
- [ ] Run all the tests of all the packages
- ...

gaborigloi:
Working on the first item, should send a PR soon. Uploaded the image https://hub.docker.com/r/gaborigloi/xs-opam/ , successfully built `xen-api-client` in it without any extra steps, by just running `make`.

gaborigloi:
Code is on this branch: https://github.com/gaborigloi/xs-opam/tree/push_docker

gaborigloi:
After running `odig odoc`, the index.html and the docs will be in ` $(opam config var prefix)/var/cache/odig/odoc`

mseri:
I have a Travis cron job running nightly to check for updates, I'll move it to this repo later today

gaborigloi:
Closing as we now have a completely different, simplified Travis setup: https://github.com/xapi-project/xs-opam/blob/master/tools/travis.sh 
We should reintroduce the required checks.
I've created #333 for uploading the docs.
We should also add automation to detect unused upstream packages, I've created https://github.com/xapi-project/xs-opam/issues/334 for that

PR #217, opened 2018-01-30
duration: update to 0.1.1 for compatibility with alcotest

mseri:


mseri:
I don't understand how this did not fail when installing `$UPSTREAM` with `xs-opam` as the base repository

gaborigloi:
Probably because we did not ask for the test dependencies, so opam did not consider that version bound.

mseri:
But we ask to install both alcotest and duration, and they have conflicting versions!

gaborigloi:
Only if we want to test duration, because alcotest is a `{test}` dep, so I think if we don't tell that to opam, it will just ignore `{test}` deps completely, including their version bounds. Which kinda makes sense, because we shouldn't be prevented from using a package if it only has conflicts in its test deps and we don't want to test it.

mseri:
Still, we have to make sure that `xs-opam` is fully consistent, tests included, because otherwise it is not reliable for the travis test builds

gaborigloi:
I've added an item to https://github.com/xapi-project/xs-opam/issues/133

PR #220, opened 2018-02-05
ezxenstore 0.1.3: add xenstore_watch

lindig:
This moves module xenstore_watch from xapi-xenopsd into this
library.


lindig:
When this gets accepted, I will update the opam file inside ezxenstore.

lindig:
This is not yet ready.

lindig:
The added libraries cause conflicts when used by components because the library is not "wrapped". this needs to be addressed first - so closing this.

PR #221, opened 2018-02-06
dlm: update to current version

mseri:


mseri:
No problem.
The latest version of my auto-update tool is made to fail if we get updates to stuff in `xs` or `xs-extra` from upstream, to enforce that our packages in `xs-opam/xs*` are always updated.

PR #223, opened 2018-02-06
xs-opam/tools: move tools in the tools subfolder and introduce opam-updates.py

mseri:
The tool is compatible with `python 2.7` and `python 3`. It requires the `requests` library to work.
```
usage: opam-updates.py [-h] [--auth AUTH] [--ignore-xs] [XS-OPAM]

xs-opam package upgrade tool

positional arguments:
  XS-OPAM      Path to the local clone of the xs-opam repository. If omitted,
               the script will do a dry-run.

optional arguments:
  -h, --help   show this help message and exit
  --auth AUTH  GitHub API authentication in the form USER:TOKEN, where TOKEN
               is a Personal Access Token (see
               https://github.com/settings/tokens). If omitted, the script
               will try to use the $HOME/.netrc file.
  --ignore-xs  Do not check if we are trying to upgrade packages in `xs` or
               `xs-extra`.
```


.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 services:
   - docker
 script:
-  - bash ./build-docker.sh
+  - bash ./tools/build-docker.sh
```
gaborigloi: Could you please update the invocation of this script in the Building the Docker Images Locally section of the readme?

mseri: Sure

README.md
```diff
@@ -199,6 +199,25 @@ env OCAML_VERSION=4.04.2 DISTRO=centos-7 BASE_REMOTE=/xs-opam COMPILE_ALL=1 ./bu
 Refer to the [`opam` manual](https://opam.ocaml.org/doc/Usage.html) for
 additional information.
 
+
+## Additional tools
+
+The tools folder contains additional utilities.
+
+- `opam-updates.py`: you can create the following file
+  ```ocaml
```
gaborigloi: This is not ocaml :P

mseri: :D

.travis.yml
mseri: Added

mseri:
The script could be easily used for point 3 in https://github.com/xapi-project/xs-opam/issues/133

tools/opam-updates.py
```diff
@@ -0,0 +1,326 @@
...
+
+def get_opam_upgrade_output():
+    """
+    Returns the output of `opam upgrade` (executed in dry mode) after
+    adding the upstream opam remote to the remotes.
+    """
+    call("opam remote add upstream https://opam.ocaml.org".split(' '))
+
+    try:
+        opam_out = check_output("opam upgrade --dry-run -y".split(' '))
```
gaborigloi: There is the `--json=file` option to dump the output in a nice json format into a file, maybe that will be easier to parse and less likely to break in the future?
```
opam upgrade --dry-run -y --json=/tmp/j
```

mseri: It was my first attempt, unfortunately it turned out that the json is completely useless. It will list all the installed packages, without any metadata to know if they were updates, upgrades, installs or reinstalls.

gaborigloi: Oh ok  - I guess the only indicator that it's a reinstall is that the version field is the same as the current one.

mseri: No, it could be an update (opam/url changed upstream but version kept unmodified) or a reinstall.

mseri: It is really useless

tools/opam-updates.py
```diff
@@ -0,0 +1,326 @@
...
+            download(session, entry["download_url"], new_path)
+
+
+def extract_upgrade(row):
+    "Extract upgraded packages from a row in the opam output."
+    name, cur, _, new = row.strip().split()[2:6]
+    return Package(name, cur, new)
+
+
+def extract_update(row):
```
gaborigloi: What is the difference between an upgraded and an updated package? Is this the "reinstall"?

mseri: upgrade: fresh new version
update: something changed upstream but the version is unchanged

tools/opam-updates.py
```diff
@@ -0,0 +1,326 @@
...
+    """
+    parser = argparse.ArgumentParser(
+        description="xs-opam package upgrade tool")
+
+    parser.add_argument("xs_opam", metavar="XS-OPAM", nargs="?", default=None,
+                        help="Path to the local clone of the xs-opam repository. "
+                        "If omitted, the script will do a dry-run.")
+
+    # In principle we don't need a default as requests can use netrc on
+    # its own, however we need to know if we can find an out or now, so
```
gaborigloi: Typo: "find an out"

tools/opam-updates.py
```diff
@@ -0,0 +1,326 @@
...
+        upstream_extra_p = os.path.join(
+            args.xs_opam, "packages", "upstream-extra")
+        xs_p = os.path.join(args.xs_opam, "packages", "xs")
+        xs_extra_p = os.path.join(args.xs_opam, "packages", "xs-extra")
+
+    out = get_opam_upgrade_output()
+
+    upgrades = [
+        extract_upgrade(row) for row in out
+        if "- upgrade" in row and "xenctrl" not in row
```
gaborigloi: Hmm, how does this match the arrow symbol:
```
  ↗  upgrade   cppo               1.6.0 to 1.6.1
```

mseri: When running in a subprocess it only uses ASCII

tools/opam-updates.py
```diff
@@ -0,0 +1,326 @@
...
+
+    with requests.Session() as session:
+        session.headers.update({"Accept": "application/vnd.github.v3+json"})
+
+        for pkg in upgrades:
+            if pkg.current_name in upstream:
+                shutil.move(
+                    pkg.current_path(upstream_p),
+                    pkg.new_path(upstream_p))
+                update_files(session, args.auth, pkg, upstream_p)
```
gaborigloi: Will this remove files from the directory that have been deleted in the new version?

mseri: There is a reason: if we had our own patches I was trying to avoid to delete them and leave them there, but I think you are right and it's better to just remove the folder and recreate it.

I'll make the change in a separate commit

PR #225, opened 2018-02-09
Add dev-repos.sh - find URLs of all repos from opam files

lindig:
This is a small tool to extract information from all opam files. It finds the URLs of all dev-repos but it would be easy to extract other information.


mseri:
Please add a note on the `Tools` in the `README`: https://github.com/xapi-project/xs-opam/blob/master/README.md#additional-tools

PR #234, opened 2018-02-28
jbuilder: update to beta18

mseri:


packages/upstream/jbuilder.1.0+beta18/opam
```diff
@@ -21,12 +18,5 @@ depends: [
   # it is only a runtime dependency so that reinstalling ocamlfind
   # doesn't resintall jbuilder
   "ocamlfind" {build}
```
lindig: This line is not present in the opam file upstream. Do we need it?

```
opam-version: "1.2"
maintainer: "opensource@janestreet.com"
authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
homepage: "https://github.com/ocaml/dune"
bug-reports: "https://github.com/ocaml/dune/issues"
dev-repo: "https://github.com/ocaml/dune.git"
license: "Apache-2.0"
build: [
  ["ocaml" "bootstrap.ml"]
  ["./boot.exe" "--subst"] {pinned}
  ["./boot.exe" "-j" jobs]
]
depends: [
  # ocamlfind is not mandatory to build packages using
  # jbuilder. However if it is present jbuilder will use it.  Making
  # it a hard-dependency avoids problems when there is a previous
  # ocamlfind in the PATH. We make it a "build" depepdency even though
  # it is only a runtime dependency so that reinstalling ocamlfind
  # doesn't resintall jbuilder
  "ocamlfind" {build}
]
available: [ ocaml-version >= "4.02.3" ]
```

mseri:
The current failure is due to the missing `bigarray` in `rrd`. The new jbuilder is stricter on dependencies, I was expecting it to fail earlier than that 😛 

Don't merge until the build is green. I am making a PR to rrd

mseri:
PR: https://github.com/xapi-project/xcp-rrd/pull/23

packages/upstream/jbuilder.1.0+beta18/opam
mseri: I did wget this from the opam file upstream. I don't understand the comment: https://github.com/ocaml/opam-repository/blob/72438cd5db7c79df9447d4662c4695440b501e5f/packages/jbuilder/jbuilder.1.0%2Bbeta18/opam#L20

lindig: Sorry. I also compared against upstream but did not notice that it has the dependency as well.

mseri: No problem :)

PR #235, opened 2018-03-02
Add lwt_log package

gaborigloi:
Prepare for the separation of the deprecated Lwt_log library: in the
future it will be split it into a completely separate lwt_log package,
which is currently a "proxy" package to make the transition smooth.


gaborigloi:
See https://github.com/ocsigen/lwt/pull/484#issuecomment-338173757 for the `lwt_log` transition plan.

PR #237, opened 2018-03-07
Update ezxenstore to v0.1.4 (Ez_venstore_watch)

lindig:
The new version of ezxenstore provides Ez_xenstore_watch that comes from
xenopsd and now can be shared across more packages. Changes in ezxenstore:

```
* 2021584 2018-03-07 CA-277850 make xenstore_watch a functor                          
* a21ead8 2018-02-02 Integrate xenctrl_uuid, xenstore_watch                           
* 831f3a8 2018-02-02 Add xenctrl_uuid from xapi-xenopsd                               
* 5565cb3 2018-02-02 Add xenstore_watch.ml from xapi-xenopds                          
```


packages/xs/ezxenstore.0.1.4/opam
```diff
@@ -12,6 +12,7 @@ depends: [
   "jbuilder" {build}
   "cmdliner"
   "logs"
+  "xenctrl"
```
gaborigloi: We should also add this to the opam file in the source repo.

mseri: Sure, after this is merged

lindig: I'll update the source repo but keep this PR as it is.

lindig: I have updated the source and moved the tag such that it is part of this PR. I've re-run the Travis check.

mseri: Sorry, I thought I was reading the change on the main PR, not on `xs-opam`

PR #244, opened 2018-03-15
xs-opam: further upstream updates

mseri:
- asn1-combinators: 0.1.3 -> 0.2.0
- bisect_ppx: 1.3.0 -> 1.3.2
- tar, tar-unix: 0.8.0 -> 0.9.0 + patch to disable -safe-string on by default
- xen-gnt: 2.2.3 -> 3.0.1
- x509: 0.5.3 -> 0.6.1
- add xen-gnt-unix 3.0.1
- add ptime 0.8.3

This requires merging https://github.com/xapi-project/rrd-transport/pull/49 and release a new xs-opam (as I believe rrd-transport is pinned, we need to double check)


mseri:
I have checked: `rrd-transport` is not pinned, it will require a new release once we release the new `xs-opam`

gaborigloi:
x509 also enabled `-safe-string`: https://github.com/mirleft/ocaml-x509/pull/89/commits/8f80a95f207f32de0b71b573021b6eb1378b9bf6 - is that a problem, do we have to add a patch as for `tar`?

mseri:
In the case of x509 it is safe because the change is not visible in the module interfaces

PR #248, opened 2018-03-22
xs-opam: auto update

mseri:
Upgrade:
- ppx_driver: v0.10.3->v0.10.4


mseri:
That's interesting, thanks. I don't think we have `odig` in the image, we should probably add it to `upstream-extra` and do it. I'll add it as soon as I have some time

gaborigloi:
That will also enable us to easily upload the docs of all the xs-opam libraries.

PR #249, opened 2018-03-23
Remove unneeded packages in upstream/ and detect them in Travis

gaborigloi:


tools/travis.sh
```diff
@@ -49,7 +52,18 @@ fi
 
 if [ "${OPAM_LINT}" = 1 ]; then
     find packages -iname opam -print | xargs -n 1 opam lint
+elif [ "${CHECK_UNUSED}" = 1 ]; then
+    opam install -y --fake $XS $(pkg xs-extra-dummy)
```
gaborigloi: This should only be used with `COMPILE_ALL=1`, otherwise it will list the packages that are only needed by xs-extra/

Dockerfile
```diff
@@ -1,14 +0,0 @@
-FROM ocaml/opam:ubuntu-16.04_ocaml-4.04.2
```
mseri: Maybe we should keep this in utils and add a job that builds `xs-opam:devel` with `xs-opam:latest + upstream + the tools and libraries installed here`

packages/xs-extra-dummy/rrdd-plugins.master/opam
```diff
@@ -0,0 +1,24 @@
...
+  "xapi-libs-transitional"
+  "xapi-stdext-std"
+  "xapi-stdext-unix"
+  "base-threads"
+  "ezxenstore"
+  "xenstore"
+  "xenops"
+  "xenctrl"
+]
+depexts: [
```
mseri: Remove this and remove the build line

mseri:
Nice!

Dockerfile
gaborigloi: Sounds good, I'll remove the commit for now, and leave this change for another PR

packages/xs-extra-dummy/rrdd-plugins.master/opam
gaborigloi: Done.

mseri:
Fails with:
```
tools/travis.sh: line 66: update_distro_packages: command not found
```

gaborigloi:
Oops typo

PR #253, opened 2018-04-11
Use the real aspcud in centos7 containers and update xs-opam

mseri:
The remote opam centos7 image was using a bash script that were forwarding the cudf requests to a service that is now down. I have recompiled the real aspcud for centos and added the rpms to the tools folder, the docker images will now remove the hack and install the real aspcud.

This also includes the auto update with:

  -  menhir                  20170607 to 20171222
  -  angstrom                0.8.0 to 0.9.0
  -  alcotest                0.8.1 to 0.8.3
  -  magic-mime              1.0.1 to 1.1.0
  -  ~~sha                     1.10 to 1.12~~
  -  mustache                3.0.0.1 to 3.0.2
  -  mirage-profile          0.7.0 to 0.8.2
  -  conduit                 1.0.3 to 1.1.0
  -  conduit-lwt             1.0.3 to 1.1.0
  -  conduit-lwt-unix        1.0.3 to 1.1.0
  -  conduit-async           1.0.3 to 1.1.0


tools/travis.sh
```diff
@@ -22,6 +23,9 @@ update_distro_packages()
     case "${DISTRO}" in
         centos*)
             sudo yum install -y epel-release
+            # remove fake aspcud and install our own rpms
+            sudo rm -rf /usr/bin/aspcud
+            sudo yum install -y tools/aspcud/*.rpm
```
gaborigloi: Do we need `tools/aspcud/aspcud_xs.tar.gz` too, or can we remove it from the PR?

lindig: Would it be better to curl/wget the RPMs rather than including them here? I understand that this adds a point of failure but adding architecture-specific RPMs into the repo is also not nice.

mseri: I left the `tar.gz` as it contains the specs and patches to generate those rpms. I preferred to leave them here to keep things simple and contained, but we can figure out an alternative

mseri: What if we put them in the xapi CI scripts repo? We only use that to wget ci related stuff after all

mseri: What if we put them in the xapi CI scripts repo? We only use that to wget ci related stuff after all

mseri:
I messed up the PR. I'll create a new one, sorry

PR #254, opened 2018-04-12
Use the real aspcud in centos7 containers and update xs-opam

mseri:
The remote opam centos7 image was using a bash script that were forwarding the cudf requests to a service that is now down. I have recompiled the real aspcud for centos and added the rpms to the tools folder, the docker images will now remove the hack and install the real aspcud.

This also includes the auto update with:

    menhir 20170607 to 20171222
    angstrom 0.8.0 to 0.9.0
    alcotest 0.8.1 to 0.8.3
    magic-mime 1.0.1 to 1.1.0
    sha 1.10 to 1.12
    mustache 3.0.0.1 to 3.0.2
    mirage-profile 0.7.0 to 0.8.2
    conduit 1.0.3 to 1.1.0
    conduit-lwt 1.0.3 to 1.1.0
    conduit-lwt-unix 1.0.3 to 1.1.0
    conduit-async 1.0.3 to 1.1.0


tools/travis.sh
```diff
@@ -22,6 +23,13 @@ update_distro_packages()
     case "${DISTRO}" in
         centos*)
             sudo yum install -y epel-release
+            # remove fake aspcud and install our own rpms
+            wget https://github.com/xapi-project/xapi-travis-scripts/raw/master/aspcud/aspcud-1.9.0-1.el7.centos.x86_64.rpm
+            wget https://github.com/xapi-project/xapi-travis-scripts/raw/master/aspcud/clasp-3.1.0-1.el7.centos.x86_64.rpm
+            wget https://github.com/xapi-project/xapi-travis-scripts/raw/master/aspcud/gringo-4.4.0-1.el7.centos.x86_64.rpm
+            sudo rm -rf /usr/bin/aspcud
+            sudo yum install -y *.rpm
+            rm *.rpm
```
gaborigloi: We have to fix up aspcud for the `check_unused` job as well - maybe we could move this into its own function, and run it from `update_distro_packages` and the `check_unused_branch`, so we don't do more work than we need to.

mseri: Ok

PR #255, opened 2018-04-12
Update sha from 1.10 to 1.12

mseri:
This breaks the name of the sha libraries, and requires a patch in `vhd-tool`


mseri:
The vhd-tool patch can be removed once we merge https://github.com/xapi-project/vhd-tool/pull/55

PR #263, opened 2018-05-01
add back alcotest-lwt, update dependencies for vhd-tool and xapi-clusterd, rpc 4.1.0->4.2.0

edwintorok:
This reverts commit 16aa11c332ae43ba2a01233bb07b3aa97db5ec3a.
We are now going to need it in xenserver too.


gaborigloi:
We'll need to add it to xapi-clusterd's opam file to avoid the unused package error.

mseri:
Also the vhd-tool ones given that we are here

gaborigloi:
Yep, could you please incorporate these changes: https://github.com/xapi-project/vhd-tool/pull/56/files#diff-86f1323c2ce87208d16429bf58348121 ? Thanks! :)

mseri:
Thanks!

mseri:
Thanks!!

edwintorok:
thanks for spotting it, I knew I must've forgotten something.

edwintorok:
I've added a commit that installs the tests too for `xs-extra-dummy`, although we should probably extend that in the future for other packages like xapi, etc. where we do run the unit tests.

PR #282, opened 2018-06-05
Release ppx_deriving_rpc & rpclib 5.4.0

gaborigloi:


gaborigloi:
This should have been `5.2.2` - I bumped the minor, but I should have incremented the bugfix instead

mseri:
It's ok, we had decided to bump the minor only if we neeed to backport bugfixes

PR #285, opened 2018-06-13
-safe-string build is green now: lets keep it that way

edwintorok:
https://travis-ci.org/xapi-project/xs-opam/builds/391797347


mseri:
If you want to cleanup further, you can also delete this block and get rid of the variable altogether: https://github.com/xapi-project/xs-opam/blob/5d23c3aeaeebc16e0aceebde6dc4bdba80f0aff5/tools/travis.sh#L65

PR #287, opened 2018-06-15
Align with upstream opam repository

lindig:
I have run a diff for the entries in `packages/upstream` against the corresponding entries in the officical opam repository and picked up the differences to align this repository as closely as possible. However, I am not sure that this would not overwrite important local modification - so I would like to ask for a careful review. In particular:

* `{build}` annotations 
* camlp4 and `preinstalled`
*  tar-unix changes


packages/upstream/camlp4.4.06+1/opam
```diff
@@ -24,6 +24,6 @@ remove: [
              "%{bin}%/camlp4prof"]
 ]
 patches: [ "f0ea53725465260556832398096cef8d3f20b49d.patch" ]
-available: [ (ocaml-version >= "4.06") & (ocaml-version < "4.07") ]
+available: [ (!preinstalled) & (ocaml-version >= "4.06") & (ocaml-version < "4.07") ]
```
mseri: You need to remove this line oherwise it will break in jenkins

packages/upstream/conf-perl.1/opam
```diff
@@ -12,5 +12,4 @@ depexts: [
   [["nixpkgs"] ["perl"]]
   [["archlinux"] ["perl"]]
   [["fedora"] ["perl-Pod-Html"]]
-  [["centos"] ["perl"]]
```
mseri: This looks like a regression to me

packages/upstream/ocb-stubblr.0.1.1/opam
```diff
@@ -15,7 +15,3 @@ depends: [
   "astring" ]
 depopts: []
 build: [ "ocaml" "pkg/pkg.ml" "build" "--pinned" "%{pinned}%" "--tests" "false" ]
-patches: [
-    "0001-bytecode-custom-needs-cclib-as-well.patch"
```
mseri: It does not seem that you are removing the parch file. Double check if we put it there before removing it

packages/upstream/tar.0.9.0/opam
```diff
@@ -14,11 +14,10 @@ build: [
...
-  "ppx_tools"         {build}
-  "ppx_cstruct"       {build}
+  "ppx_tools"
+  "ppx_cstruct"
   "cstruct"           {>= "1.9.0"}
   "re"
   "result"
 ]
 available: [ ocaml-version >= "4.03.0" ]
-patches: [ "0001-Do-not-always-build-with-safe-string-this-can-break-.patch" ]
```
edwintorok: You may want to delete this file if we are not using it. Would probably be useful to do a jenkins build, but I think that aside from oxenstored we are safe-string compatible now so it should be fine.

packages/upstream/ocb-stubblr.0.1.1/opam
lindig: The patch file was introduced by you and as far as I can tell the change it introduces is not current version. So we better keep it.

https://github.com/pqwy/ocb-stubblr/pull/11

packages/upstream/tar.0.9.0/opam
lindig: So let's keep this until we switch the compiler. I'll put it back in.

packages/upstream/tar-unix.0.9.0/opam
```diff
@@ -22,4 +22,3 @@ depends: [
   "lwt"
 ]
 available: [ ocaml-version >= "4.03.0" ]
-patches: [ "0001-Do-not-always-build-with-safe-string-this-can-break-.patch" ]
```
edwintorok: Should you put the patch back in here as well?

mseri: :O well spotted. I was tricked by the other re-added patch :P

mseri: No, I take it back. I see it in the latest commit

lindig: I have added it back - see the last commit. Well spotted.

PR #291, opened 2018-06-25
Update rpclib, rpclib-async, rpclib-ppx, ppx_deriving_rpc to 5.8.0

gaborigloi:


lindig:
This looks good but let's wait for a Travis build.

packages/upstream/rpclib-async.5.8.0/url
```diff
@@ -1,2 +1,2 @@
 archive: "https://github.com/mirage/ocaml-rpc/archive/v5.1.0.tar.gz"
-checksum: "97f526cf3c8c8fca033df1175d4d23df"
+checksum: "4dc2050aeba98fe3a538377b17205348"
```
edwintorok: Why would a checksum of an already released tarball change?
Did you mean to update the version number here?

packages/upstream/rpclib-lwt.5.8.0/url
```diff
@@ -1,2 +1,2 @@
 archive: "https://github.com/mirage/ocaml-rpc/archive/v5.1.0.tar.gz"
-checksum: "97f526cf3c8c8fca033df1175d4d23df"
+checksum: "4dc2050aeba98fe3a538377b17205348"
```
edwintorok: Same question as above.

gaborigloi: Well-spotted, I forgot to update those urls.

packages/upstream/ppx_deriving_rpc.5.8.0/url
```diff
@@ -0,0 +1,2 @@
+archive: "https://github.com/mirage/ocaml-rpc/archive/v5.8.0.tar.gz"
+checksum: "3ba9648e65197641a38ccfedb9135b70"
```
edwintorok: The checksum is wrong here.
Can't we use some automated tool to update the version here? (maybe that daily script that we used to run that checks for new upstream versions?) Looks like it is very error prone to do this manually.

mseri: We are not yet publishing ocaml-rpc upstream. The version there is way outdated. After my idl changes are merged I'll release it upstream and will try to keep the release synced.

mseri: It is probably possible to use dune-release (or a slightly patched version) to do it though

gaborigloi: I was thinking of updating our python script in xs-opam so that it can look at a source repo and update it to the latest release in the source repo of the package

edwintorok:
Note that `xcp-networkd` fails to build and will need a patch like this https://github.com/xapi-project/xcp-networkd/commit/858ee911e65e3a4d47315de6239e40b3b14496ae

gaborigloi:
The PR is ready: https://github.com/xapi-project/xcp-networkd/pull/144 and it builds with that change

lindig:
Let's merge xcp-networkd first and run Travis again here.

lindig:
There is one failure but is this expected because it is using outdated RPMs?

mseri:
No, `xs-opam` does not use rpms. From a brief look `xapi-plugin` has incorrect dependencies, `unix` (and the now dead `rpclib.unix` are not needed). I wonder if anything uses that code though

edwintorok:
https://github.com/xapi-project/ocaml-xapi-plugin/pull/2 the dependency was not needed, it builds without.

mseri:
I actually don't think that `xapi-plugin` is used at all... Have a look, you may be able to archive it and remove it from `xs-opam`

gaborigloi:
I don't think it's used.

gaborigloi:
I think we can merge this now?

edwintorok:
Hmm why did the build here succeed when ocaml-xcp-idl's tests didn't even build, is it not running the  tests?

mseri:
No, we dropped it because with the tests was taking hours (and requiring a lot more dependencies). The reason is that opam 1 will run for each package all tests of all the dependencies including upstream. This is not fixable until opam 2 is released

edwintorok:
Build on xcp-idl PR failed due to missing `diet` package, but that is not a problem introduced by this PR. Lets merge this and I'll open another PR to add the diet package.

PR #292, opened 2018-06-26
Catch up with upstream changes

lindig:
Packages in upstream/ moved on or details were updated. Here we are catching up with most of these changes.


lindig:
Ping @mseri.

PR #293, opened 2018-06-26
Add missing package diet: required by Travis on xcp-idl due to mirage-block-unix

edwintorok:
See build failure on https://github.com/xapi-project/xcp-idl/pull/231

### DIFF
--- a/packages/upstream-extra/diet/diet.0.1/descr
+++ b/packages/upstream-extra/diet/diet.0.1/descr
@@ -1,0 +1,6 @@
+Discrete Interval Encoding Trees
+
+This is based on the
+[Functional Pearls: Diets for Fat Sets](https://web.engr.oregonstate.edu/~erwig/papers/Diet_JFP98.pdf)
+by Martin Erwig.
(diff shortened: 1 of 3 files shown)


gaborigloi:
Maybe we could put this into `packages/upstream-extra` instead.
Or we could make a new directory for test dependencies, `packages/upstream-testing`

lindig:
If we create a new directory we will have to be careful about our existing scripts that would miss it.

mseri:
If it is used only by tests, it shuold go in `packages/upstream-extra`, otherwise in `packages/upstream`. In any case `packages/diet` is wrong

lindig:
I am not so happy about this PR breaking the existing directory structure and creating a special case. I think `upstream-extra` would work. These packages are available for Travis but are not packaged into the RPMs we are building. Only `upstream` and  `xs` are packaged.

edwintorok:
Good points, I've moved the package.

edwintorok:
Build passed

PR #295, opened 2018-06-26
Check missing test-only dependencies and add missing ppx_driver and conf-ncurses

edwintorok:
Travis on xcp-idl now failed with missing `ppx_driver` dependency.
Add a test in xs-opam that checks for missing test-only dependencies, so
we can find them all more easily.

This command can be used locally to quickly check this (it won't actually build the packages just check that they are installable)
```
BASE_REMOTE=/xs-opam CHECK_DEPENDENCIES=1 COMPILE_ALL=1 OCAML_VERSION=4.06.0 DISTRO=centos-7 tools/build-docker.sh
```


mseri:
I think `check_unused` and `check_dependencies` should be two separate tests. The first is to make sure that the packages in `xs` are only the ones really needed by xapi, the other is to make sure that we can run the tests in travis

edwintorok:
We've run them as one step to conserve Travis resources, most of the overhead is in starting a new build step, once running we can easily test for both. I agree that it makes it harder to see or fix just one of the problems, and the precedence is not correct: we care more about tests running than having unused packages.

edwintorok:
Maybe we could use build stages: put the one checking for (test) dependencies as first stage, and skipping all other builds if that one fails, and have the other ones all part of second stage: https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions

mseri:
Yes and no. Bloating the installed xs-opam in Jenkins will slow down the build

mseri:
I like your suggestion

edwintorok:
Still testing it, looks like it only picked up one build stage.

edwintorok:
Looks like it correctly picked up both stages now, had to separate the two `env` lines

PR #296, opened 2018-07-05
CA-292641: Update xapi-stdext-pervasives to 1.1.0

gaborigloi:
Changes:
* CA-292641: Use Logs to log cleanup exn instead of shadowing the
  original one with it


gaborigloi:
Because the stdext repo contains multiple opam packages, with different versions.

gaborigloi:
Is this good to merge?

PR #297, opened 2018-07-06
Catch up with upstream updates

lindig:
- uri: 1.9.6->1.9.7
- ezjsonm: 0.5.0->0.6.0
- travis-opam: 1.0.3->1.2.0


lindig:
Good catch. I've fixed the commit message and force-pushed the change.

PR #298, opened 2018-07-11
Catch up with upstream package updates

lindig:
* add dune
* Update jbuilder
* Update menhir
* Update mirage-flow, mirage-flow-lwt


mseri:
Already unsing dune!!

PR #302, opened 2018-07-25
vhd-format: add 2 bugfix patches

lindig:
This adds two patches that correct bugs in existing patches:

* 0003-Fix-string-escaping-in-code-generator.patch

  A call to String.escaped was lost when making the
  code compatible with immutable strings

* 0004-Fix-detection-of-unused-entries.patch

  A comparison for 0xffff_ffff to detect unused table entries
  was not effective.


lindig:
Yes, BVT and BST passed (minus one known, unrelated issue). I am confident that the new patches fix the previous problem.

PR #309, opened 2018-08-15
Update vhd-format to 0.10.0

lindig:
* XSI-19: locate the footer in a VHD file based on the total size of
  the file.
* Patches have been upstreamed and are now removed from the Opam entry,


lindig:
This introduces a breaking change for vhd-format (ocaml-vhd). Currently vhd-tool fails to compile until we update it. However, vhd-tool is not part of xs-opam but just a client. vhd-format introduced a new dependency: lwt_ppx.

lindig:
If we update vhd-tool the internal build on master will fail until we release xs-opam. But some breakage is unavoidable.

PR #310, opened 2018-08-18
Fix GitHub archive URLs

lindig:
GitHub no longer accepts archive URLs that contain an arbitrary file
name. Instead, it can only contain the tag it refers to. This commit
updates the URL files to conform to this. Furthermore, it adds a newline
to the last line in the file if it was missing.


mseri:
It would be easier to review if the changes to the archive line were in a separate community. Lots of the changes are only about the end line at the end of the file. It seems fine though

lindig:
I've split this into two commits.

lindig:
This might not be required because GitHub seems to serve again the existing URLs.

PR #109, opened 2017-07-07
Update README

mseri:
The updated README removes a now unnecessary step 
and introduces the answers to some common questions
that have been asked multiple times in the last weeks.

Signed-off-by: Marcello Seri <marcello.seri@citrix.com>

PR #312, opened 2018-08-30
Catch up with upstream changes

lindig:
This PR takes into account changes in upstream libraries. In particular it updates the ppx_tools_versioned package which introduces unwanted dependencies into generated in version 5.2.


packages/upstream/x509.0.6.1+1/files/0001-x509-add-support-for-ppx_sexp_conv.v0.11.0.patch
```diff
@@ -1,141 +0,0 @@
-From 9b515187d79b4dbdfa23091e43ba5ba00a8b620b Mon Sep 17 00:00:00 2001
```
gaborigloi: I think it's okay to remove this:
https://github.com/mirleft/ocaml-x509/blob/master/CHANGES.md
> 0.6.2 (2018-08-24)

>    compatibility with ppx_sexp_conv >v0.11.0 (#109), required for 4.07.0

PR #313, opened 2018-09-03
Remove upstream-extra/depext.1.0.5

lindig:
It's no longer needed, we are using opam-depext 1.1.2.


edwintorok:
Looks like Travis is still looking for it: https://travis-ci.org/xapi-project/xcp-idl/builds/423999509?utm_source=github_status&utm_medium=notification

lindig:
I didn't like that we had two versions in there - will look into it.

PR #315, opened 2018-09-04
Use opam2 as base for docker containers

edwintorok:
Opam2 uses 4.06.1 as compiler by default, and doesn't have per compiler
version containers for all distros anymore, see:
https://hub.docker.com/r/ocaml/opam2/tags/

The preferred way to use another ocaml version is to use opam switch
create:
https://github.com/ocaml/ocaml-ci-scripts/blob/master/.travis-docker.sh#L62

Since we are using 4.06.1 now doing so would be redundant, but when we
switch to 4.07 this would unfortunately slow down our build times as
we'd need to build the compiler itself too.

Lets wait and see whether the build succeeds on Travis.


lindig:
See branch `407` for a simplified setup with OCaml 4.07

edwintorok:
Thanks, that branch had some useful opam2 commands. I like the xs-toolstack dummy package and a single-purpose travis.sh, our current travis.sh tries to do too much. The dependency checking, etc. should probably be separate scripts that contain only the stuff needed for that particular action.
I'll see if I can port some commits from that branch here (except the ones that change the repository format).

lindig:
Please merge this

PR #316, opened 2018-09-12
Simplify Travis setup

lindig:
This drastically simplifies the Travis setup for this repository because I saw builds frequently break and was no longer able to navigate the maze of conditions, environment variables, and shell scripts. 

* There is only one build which is triggered by the virtual package `xs-toolstack`
* Everything is built in a container that is provided by `ocaml/opam` and determines the compiler
* Everything is built using Opam 2
* A build takes about 15 minutes


packages/xs-extra-dummy/xs-toolstack.master/opam
```diff
@@ -0,0 +1,56 @@
+# vim: set et:
+#
+
+opam-version:   "1.2"
+name:           "xs-toolstack"
+maintainer:     "Christian Lindig <christian.lindig@citrix>"
```
edwintorok: missing .com?

tools/travis.sh
```diff
@@ -1,91 +1,23 @@
...
-    # Fail if there are unused packages in $UPSTREAM:
-    AVAILABLE=$(pkg upstream | sed 's/\..*$//' | sort -u)
-    INSTALLED=$(opam list | grep -v \# | cut -d' ' -f1 | sort -u)
-    UNNEEDED=$(comm -23 <(echo "$AVAILABLE") <(echo "$INSTALLED"))
-    if [ -n "$UNNEEDED" ]; then
-        echo Unused packages in upstream/: $UNNEEDED
-        exit 1
-    fi
-elif [ "${CHECK_TEST_DEPENDENCIES}" = 1 ]; then
-    opam install -t -y --fake $XS
```
edwintorok: We could keep these dependency checks, but in a separate script. I agree that there were too many of them. Perhaps just one script and Travis matrix entry that does all the dependency, and unused package checks?

Or we could have these checks done in the `tools/opam-updates.py` script that we run from a cronjob, what do you think?

About the docker container: I am using it here: https://github.com/edwintorok/xs-opam-dev, but if it isn't generally useful I could move building the base of it to my repo as well. (although not sure about `tools/opam-updates.py` it might need the container). Even if we keep the docker container we could use a simplified form of it and build just one combination (say CentOS-7 cause thats what we use internally) instead of being generic for any distribution.

tools/travis.sh
```diff
@@ -1,91 +1,23 @@
...
+# 
 
-set -ex
+set -e
 
-# override xenctrl.dummy to have a full-fledged opam repository
-if [ ! "${COMPILE_ALL}" = 1 ]; then
-    sudo sh tools/into_repo.sh
-fi
+IMG='ocaml/opam2:debian-9-ocaml-4.06'
```
edwintorok: Comment says 4.07, image says 4.06, which one did you intend to use?

tools/travis.sh
lindig: I understand that `opam-updates` uses the container but I would propose we add building the container back in later.

edwintorok:
Merging, because the current Travis build is broken, and the one on this PR isn't.

PR #318, opened 2018-09-13
Update Opam files for xenopsd, xapi

lindig:
This updates opam files such that they are aligned with the ones in the code repositories.

* xen-api (xapi)
* xenopsd


packages/xs-extra/xapi-xenopsd-xenlight.master/opam
```diff
@@ -7,28 +7,11 @@ dev-repo: "https://github.com/xapi-project/xenopsd.git"
 bug-reports: "https://github.com/xapi-project/xenopsd/issues"
```
gaborigloi: Can we get rid of this file as we removed the xenlight backend from xenopsd?

PR #324, opened 2018-09-20
Add bigstringaf required by angstrom

lindig:
Also make sure we build angstrom which is required by xapi-clusterd but
which is not built by itself.


packages/xs-extra-dummy/xs-toolstack.master/opam
```diff
@@ -8,6 +8,7 @@ authors:        "Christian Lindig <christian.lindig@citrix.com>"
 homepage:       "https://github.com/xapi-project/xs-opam"
 license:        "BSD"
 depends: [
+  "angstrom"
```
gaborigloi: Could we instead depend on the `xapi-clusterd` package in `xs-extra-dummy`?

edwintorok: It is already there, maybe we can add all the dummy packages as dependency? @gaborigloi  says they are installable because they have no build rule, just purely dependencies.
These are the dummy packages: rrdd-plugins, xapi-clusterd-test, xapi-clusterd, xenctrl

lindig: No - this packages has no URL because it is private and can't be installed (I believe).

gaborigloi: I have tried installing it with `opam install xapi-clusterd` (with opam 2) and it worked for me.

edwintorok: https://github.com/xapi-project/xs-opam/blob/master/packages/xs-extra-dummy/xapi-clusterd.master/opam

edwintorok: We've added a dummy opam package for each of these, containing just the dependencies, and we deleted the other fields.

lindig: I have added more dummy packages.

PR #321, opened 2018-09-18
Update xen-api-client, message-switch and xapi-idl opam files

edwintorok:
This is to reduce our dependency on "rpc", which is not really correct, since we depend on rpclib in fact.
Still more to go though, so we could keep this PR open until we gathered most of them:
```
packages/xs-extra-dummy/xapi-clusterd-test.master/opam:  "rpc" { >= "1.9.53" }
packages/xs-extra-dummy/xapi-clusterd.master/opam:  "rpc" { >= "1.9.53" }
packages/xs-extra/forkexec.master/opam:  "rpc" {>= "1.9.51"}
packages/xs-extra/http-svr.master/opam:  "rpc"
packages/xs-extra/rrdd-plugin.master/opam:  "rpc"
packages/xs-extra/xapi-networkd.master/opam:  "rpc"
packages/xs-extra/xapi-plugin.master/opam:  "rpc"
packages/xs-extra/xapi-rrdd.master/opam:  "rpc"
packages/xs-extra/xapi-squeezed.master/opam:  "rpc"
packages/xs-extra/xapi-storage-cli.master/opam:  "rpc"
packages/xs-extra/xapi-storage.master/opam:  "rpc"
packages/xs-extra/xapi-tapctl.master/opam:  "rpc"
packages/xs/rrd.1.3.0/opam:  ( "rpc" {>= "1.9.51" & < "5.0.0"}
packages/xs/xapi-backtrace.0.6/opam:  "rpc" {>= "1.9.51"
```

PR #325, opened 2018-09-20
Sync opam files with code repos

lindig:
This updates opam files with versions that are in the corresponding code
repositories. In a few cases this is a regression where the maintainer
is updated but it makes the repositories the source of truth, as it
should be.


edwintorok:
Thanks, if the build works, I think this would obsolete https://github.com/xapi-project/xs-opam/pull/321

PR #327, opened 2018-09-24
Update Travis setup

lindig:
This is a hack but Travis works again at least for the upstream libraries. It does not try to compile clients of xs-opam.


lindig:
I'm not sure about the compiler variants. I'll trim this list.

PR #328, opened 2018-09-26
Catch up with upstream

lindig:
This updates packages in upstream/. This was constructed by updating URLs and checksums manually but it had to reuse the existing Opam files because the upstream Opam repository no longer provides Opam meta data in Opam 1 format.


mseri:
With the new cohttp you should be able to move to lwt 4.0.0

PR #329, opened 2018-10-09
make depext work again, various improvements

lindig:
Using this repository as a base repository so far only worked for opam install but failed for opam depext for reasons that are not entirely clear. This series of commits tried to fix it while also making sure it still works for building RPMs from an archive of this repository.

* The "all" target is more useful now: it builds a docker container which can be used for development
* packages/ocaml/base-* was in Opam 1 format - so updated to Opam 2
* added ocam-base-compiler - this is required to make depext work although this is not obvious from the dependencies
* in tools/travis.sh, use "opam depext" again
* When building a tarball, don't cache the OCaml compiler
* Accidentally already pushed with a prior commit: a patch for Core's time function

```
 Makefile                                           |  5 +++-
 packages/ocaml/base-bigarray.base/descr            |  1 -
 packages/ocaml/base-bigarray.base/opam             |  9 +++---
 packages/ocaml/base-bytes.base/descr               |  1 -
 packages/ocaml/base-bytes.base/opam                |  9 +++---
 packages/ocaml/base-threads.base/descr             |  1 -
 packages/ocaml/base-threads.base/opam              |  9 +++---
 packages/ocaml/base-unix.base/descr                |  1 -
 packages/ocaml/base-unix.base/opam                 |  9 +++---
 packages/ocaml/ocaml-base-compiler.4.06.1/opam     | 33 ++++++++++++++++++++++
 .../upstream/core.v0.11.3/files/ca-297060.patch    | 12 ++++++++
 packages/upstream/core.v0.11.3/opam                |  1 +
 tools/Dockerfile                                   | 15 ++++++++++
 tools/travis.sh                                    |  1 +
```

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -5,11 +5,12 @@ MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
 RUN sudo apt-get install -qq -yy \
   debianutils hwdata libdlm-dev libffi-dev libgmp-dev libnl-3-200   \
   libnl-route-3-200 libpam-dev libpci-dev libssl-dev libsystemd-dev \
-  libxen-dev linux-libc-dev m4 perl pkg-config python uuid-dev
+  libxen-dev linux-libc-dev m4 perl pkg-config python uuid-dev \
+  && sudo apt-get clean
 
-COPY . /mnt
+COPY . /tmp/xs-opam
 
-RUN opam repo remove --all default
-RUN opam repo add xs-opam file:///mnt
-RUN opam depext -y xs-toolstack
-RUN opam install -j 4 xs-toolstack
+RUN opam repo remove --all default \
+ && opam repo add xs-opam file:///tmp/xs-opam \
+ && opam depext -y xs-toolstack \
+ && opam install -j 4 xs-toolstack
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,0 +1,15 @@
+
+FROM        ocaml/opam2:debian-9-ocaml-4.06
+MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
+
+RUN sudo apt-get install -qq -yy \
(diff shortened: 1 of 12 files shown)


tools/Dockerfile
```diff
@@ -0,0 +1,15 @@
+
+FROM        ocaml/opam2:debian-9-ocaml-4.06
+MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
+
+RUN sudo apt-get install -qq -yy \
+  debianutils hwdata libdlm-dev libffi-dev libgmp-dev libnl-3-200   \
+  libnl-route-3-200 libpam-dev libpci-dev libssl-dev libsystemd-dev \
+  libxen-dev linux-libc-dev m4 perl pkg-config python uuid-dev
```
edwintorok: If you want to make the image smaller you could do a `&& sudo apt-get clean` in the same RUN line (doing it on a separate RUN line would not have any benefits as the previous RUN line would've already created a layer in the docker image)

edwintorok: Wouldn't the depext below already take care of installing most of the above though? Although I can see the benefit in caching even if it means having duplicate info, so lets keep it for now.

tools/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY . /mnt
+
+RUN opam repo remove --all default
+RUN opam repo add xs-opam file:///mnt
+RUN opam depext -y xs-toolstack
```
edwintorok: These 4 could also be on the same line, if anything changes in `.` then docker will need to recreate every layer from the COPY line, so the ones below it wouldn't be cached, and having it on a single line could mean a smaller image.
If we want we could also rm opam's cache like done in the spec file to make the container smaller.

tools/Dockerfile
lindig: The apt-get could be removed. It is is from the time when `depext` did not work. I'll add an `apt-get clean`.

PR #331, opened 2018-10-19
Sync cohttp opam files with upstream repo

gaborigloi:
This adds an important version bound on uri, which prevents compilation
failures with uri 2.0.0, which is the latest upstream version.


mseri:
cohttp-1.2.0 has been released

PR #335, opened 2018-10-29
Ported xen-api-sdk to dune.

kc284:
This pairs with https://github.com/xapi-project/xen-api-sdk/pull/154


kc284:
Will merge after travis finishes.

lindig:
I believe either GitHub or Travis have networking issues. Several runs failed with problems to fetch the sources.

PR #336, opened 2018-10-30
ssl.0.5.6 -> ssl.0.5.7

edwintorok:
ssl 0.5.6 failed to get downloaded from github due to an extra 'v' in
the URL.
Update to 0.5.7 which includes the patch that we locally applied to 0.5.6.


edwintorok:
It appears that upstream deleted the v0.5.6 tag for consistency, I have the tag in an old checkout, but don't have it in a fresh clone.

PR #338, opened 2018-11-01
Sync opam files from component repositories

lindig:
Use the same opam files as they are in the repositories of these
packages:

* packages/xs-extra/nbd-tool.master/opam
* packages/xs-extra/rrd-transport.master/opam
* packages/xs-extra/xapi-storage-cli.master/opam
* packages/xs-extra/xapi-storage.master/opam
* packages/xs-extra/xen-api-sdk.master/opam

### DIFF
--- a/packages/xs-extra/xapi-storage.master/opam
+++ b/packages/xs-extra/xapi-storage.master/opam
@@ -13,6 +13,10 @@ depends: [
   "ocaml"
   "jbuilder" {build}
   "ounit" {with-test}
+  "alcotest" {with-test}
+  "lwt" {with-test}
+  "rpclib" {with-test}
+  "ppx_deriving_rpc" {with-test}
   "rpc"
   "xmlm"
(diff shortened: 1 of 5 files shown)


packages/xs-extra/xapi-storage.master/opam
```diff
@@ -13,6 +13,10 @@ depends: [
   "ocaml"
   "jbuilder" {build}
   "ounit" {with-test}
+  "alcotest" {with-test}
+  "lwt" {with-test}
+  "rpclib" {with-test}
+  "ppx_deriving_rpc" {with-test}
```
edwintorok: This one looks a bit odd, why would this be a test-only dependency? If the build passes we can merge though

lindig: I used Dune to query the dependencies and they only show up for `@runtest`

PR #348, opened 2018-11-19
Various updates

lindig:
* Remove obsolete depext.1.0.5
* Add opam-ed - required by OCaml CI scripts
* Add deps2dot for deps graph
* Update upstream packages
** dune
** menhir
** mirage-random


tools/cleanup.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-exec opam admin filter --dry-run --resolve=xs-toolstack --resolve upstream-extra-dummy --resolve depext.transition --resolve depext.1.0.5 --or --with-test
+exec opam admin filter --dry-run --resolve=xs-toolstack --resolve upstream-extra-dummy --resolve depext.transition --resolve opam-ed.0.1 --or --with-test
```
gaborigloi: What does this script do?

gaborigloi: We should document it.

lindig: It detects packages that are not referenced.

PR #352, opened 2018-11-26
CP29693 - xenopsd - remove patch for xeneventchn

lindig:
This removes a patch that we introduced only to make xenopsd work on
Travis but which was never used in production. The patch removed a
dependency on xeneventchn.


edwintorok:
We would expect the build to fail here, and once we merge the xenopsd PR it should pass.

PR #357, opened 2018-12-03
nbd-lwt-unix: add lower bound on nbd package to ensure compatibility

gaborigloi:
This is to ensure that we will compile nbd-lwt-unix with a compatible
version of the nbd library. Nbd 4.0.0 has introduced some new values
that this version of nbd-lwt-unix relies on, it does not work with nbd
3.X.


gaborigloi:
This change is required to get the dev-vm to work, currently one ticket is blocked by this.

PR #372, opened 2019-01-31
Run tests in Travis build

edwintorok:
See https://github.com/xapi-project/xapi-storage/pull/97#issuecomment-459322492
Would be useful to try and run these in Travis if they do not take a long time to run (but of course we'll also have to fix the tests before merging this PR)


lindig:
Closing this as it needs more work before we can merge it.

PR #373, opened 2019-03-14
Release stdext 4.5.0

lindig:
Also rename all aliases to use the same version number as the underlying
code base.


lindig:
Sorry - this can't be merged because I missed a previous commit around stdext due to an outdated master branch.

PR #374, opened 2019-03-15
CP-30756: Update upstream dependencies with minor updates

psafont:


tools/Dockerfile
```diff
@@ -1,11 +1,10 @@
...
 
-COPY . /tmp/xs-opam
+COPY --chown=opam:opam . /tmp/xs-opam
 
 RUN sudo apt-get update
 RUN opam repo remove --all default \
  && opam repo add xs-opam file:///tmp/xs-opam \
  && opam depext -y xs-toolstack \
- && opam install -j 4 xs-toolstack
+ && opam install xs-toolstack
```
edwintorok: Did you intend to remove the `-j` here?

psafont: Actually, I changed it but didn't intend to commit it. In my machine it uses `11` as default, from the docs:

```
       -j JOBS, --jobs=JOBS
           Set the maximal number of concurrent jobs to use. The default value
           is calculated from the number of cores. You can also set it using
           the $OPAMJOBS environment variable.
```
[default](https://github.com/ocaml/opam/blob/9778ada9360b9e32ff0ae025999d6f77831dc211/src/state/opamStateConfig.ml#L35)

edwintorok: Ok then removing it is probably better since it allows more parallelism

PR #383, opened 2019-04-29
Prepare for xxhash external dependency

lindig:
Xapi will soon gain xxhash as a new external dependency. it is not yet available in the current distribution we use for builds on Travis. Therefore:

* record the dependency in xapi's opam file
* use debian-unstable for Travis

### DIFF
--- a/tools/travis.sh
+++ b/tools/travis.sh
@@ -6,7 +6,7 @@
 
 set -e
 
-IMG='ocaml/opam2:debian-9-ocaml-4.07'
+IMG='ocaml/opam2:debian-unstable'
 
 docker pull $IMG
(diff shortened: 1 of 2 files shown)


tools/travis.sh
```diff
@@ -6,7 +6,7 @@
 
 set -e
 
-IMG='ocaml/opam2:debian-9-ocaml-4.07'
+IMG='ocaml/opam2:debian-unstable'
```
edwintorok: I think we'll need to add an `opam switch 4.07` below, the default for this image seems to be 4.06

lindig: When I watched this build, it used 4.07

lindig: I've added it.

PR #387, opened 2019-05-15
Update cstruct to 4.0 and its dependents

psafont:
`nocrypto` has been changed to have the same patches as the default opam repository. The only difference is that the flag `"--accelerate" "false"` was kept.

`cstruct-sexplib` has been added as it's needed by `x509.0.6.3.` It looks like it's next version is going to break API and also drop the dependency.


packages/upstream/nocrypto.0.5.4+2/opam
```diff
@@ -35,13 +34,19 @@ depends: [
...
 ]
+
 patches: [
-    "0001-Update-to-support-latest-cstruct-and-ppx_sexp_conv-a.patch"
+    "0001-add-missing-runtime-dependencies-in-_tags.patch"
+    "0002-add-ppx_sexp_conv-as-a-runtime-dependency-in-the-pac.patch"
+    "0003-Auto-detect-ppx_sexp_conv-runtime-library.patch"
+    "0004-pack-package-workaround-ocamlbuild-272.patch"
+    "0005-use-modern-cstruct-findlib.patch"
+    "0006-explicit-dependency-on-sexplib.patch"
```
edwintorok: This package has accumulated quite a few packaging patches, looks like upstream hasn't made a release (or merged PRs) in quite a while

psafont: While the latest release was done in 2017, master was last updated on February. The PR for porting the build system to dune got stuck on early April: https://github.com/mirleft/ocaml-nocrypto/pull/158

PR #390, opened 2019-06-04
Update stdext to 4.7.0

lindig:
This is part of the update of base64. 


lindig:
There are several packages referring to the same github repo. I have updated them to 4.7.0 as well.

PR #396, opened 2019-07-24
Drop rpc in faviour of rpclib

edwintorok:
'rpc' is deprecated and we use 'rpclib' everywhere in the code already. Keeping 'rpc' as a dependency prevents updating other libraries (e.g. core and sync), which in turn prevents us from building on 4.08 due to old versions.

After the build passes here (it already did locally) I'll open PRs for individual repos to update their opam files too.


psafont:
Why is `rpclib` being added an option instead of replacing `rpc`?

Were all the packages using `"ppx_deriving_rpc"` imported through `rpc`?

The opam files will also need to be updated in the corresponding repositories.

edwintorok:
We can remove `rpc` completely now, it would work with both I think, but since we no longer even have `rpc` in this repo we can drop it.

PR #400, opened 2019-08-01
Add a build that also runs the tests

edwintorok:


tools/travis.sh
```diff
@@ -9,15 +9,21 @@ set -e
 IMG='ocaml/opam2:debian-unstable'
 
 docker pull $IMG
-docker run --rm -iv $PWD:/mnt $IMG <<'EOF'
-set -e
+if [ "${OPAMWITHTEST}" = "true" ]; then
+    # opam 2.x only tests packages listed on the cmdline, unlike opam 1.x
+    INSTALL="opam list ${RECURSIVE} -s --required-by xs-toolstack | xargs opam install -j $(getconf _NPROCESSORS_ONLN) -t"
```
lindig: The `-j` flag is no longer required as `opam` does it by default.
```
       -j JOBS, --jobs=JOBS
           Set the maximal number of concurrent jobs to use. The default value
           is calculated from the number of cores. You can also set it using
           the $OPAMJOBS environment variable.
```

.travis.yml
```diff
@@ -4,3 +4,11 @@ services:
   - docker
 script:
   - bash tools/travis.sh
+matrix:
+  include:
+    - env: OPAMWITHTEST=false
+    - env: OPAMWITHTEST=true
+    - env: OPAMWITHTEST=true RECURSIVE=--recursive
+  fast_finish: true
```
lindig: I would move this line under `matrix:` to emphasize that this is valid for all entries in the matrix.

PR #405, opened 2019-08-12
xs-extra: update opam files from repos

psafont:
Some files have not been pulled because of the rpc to rpclib migration, I don't know if the "ppx_deriving_rpc" are safe to remove.

### DIFF
--- a/packages/xs-extra/http-svr.master/opam
+++ b/packages/xs-extra/http-svr.master/opam
@@ -14,7 +14,6 @@ depends: [
   "dune" {build}
   "astring"
   "base64" {>= "3.1.0"}
-  "ppx_deriving_rpc"
   "rpclib"
   "sha"
   "stunnel"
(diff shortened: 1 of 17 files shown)


packages/xs-extra/xapi-plugin.master/opam
```diff
@@ -15,7 +15,3 @@ depends: [
   "rpclib"
 ]
 synopsis: "Library to simplify writing xapi plugins in OCaml"
-url {
-  src:
-    "https://github.com/xapi-project/ocaml-xapi-plugin/archive/master/master.tar.gz"
```
edwintorok: I think we need the URL link in xs-opam

packages/xs-extra/xapi-storage.master/opam
```diff
@@ -26,7 +26,3 @@ depexts: [
   ["python"] {os-distribution = "ubuntu"}
 ]
 synopsis: "Code and documentation generator for the Xapi storage interface"
-url {
-  src:
-    "https://github.com/xapi-project/xapi-storage/archive/master/master.tar.gz"
```
edwintorok: Same here: in the repo we don't but in xs-opam we need the URL link?

PR #406, opened 2019-08-12
CP-32055: Upstream updates

psafont:


packages/upstream/ssl.0.5.9/opam
```diff
@@ -9,16 +9,16 @@ build: [
...
-  "dune" {build & >= "1.2.1"}
+  "dune" {>= "1.2.1"}
   "base-unix"
   "conf-openssl"
 ]
 synopsis: "Bindings for OpenSSL"
 authors: "Samuel Mimram <samuel.mimram@ens-lyon.org>"
 url {
-  src: "https://github.com/savonet/ocaml-ssl/archive/0.5.7.tar.gz"
+  src: "https://github.com/savonet/ocaml-ssl/archive/0.5.9.tar.gz"
```
edwintorok: Lets defer this `ssl` update for now: it enables certificate checking by default which require updating a lot of tests and code on our side to deal with it.
We should update everything else, and leave this one last as a separate PR, branch.

packages/upstream/topkg.1.0.1/opam
```diff
@@ -41,9 +41,8 @@ Topkg-care is distributed under the ISC license it depends on
 [logs]: http://erratique.ch/software/logs
 [bos]: http://erratique.ch/software/bos
 [cmdliner]: http://erratique.ch/software/cmdliner
-[webbrowser]: http://erratique.ch/software/webbrowser
-"""
+[webbrowser]: http://erratique.ch/software/webbrowser"""
 url {
-src: "http://erratique.ch/software/topkg/releases/topkg-1.0.0.tbz"
-checksum: "md5=e3d76bda06bf68cb5853caf6627da603"
+  src: "http://erratique.ch/software/topkg/releases/topkg-0.9.1.tbz"
```
edwintorok: You downgraded?

psafont: picked the wrong version from opam

edwintorok:
I think you need to update core and all its dependencies to v0.12.3

psafont:
Jane street dependencies can't be updated until we start using ssl 0.5.9.

This is because conduit has to be updated in lockstep with Jane libs to 2.0.0 and conduit-lwt-unix condlict with ssl lower than 0.5.9.

psafont:
If anybody wants to review the last commit, I recommend downloading the branch and lowering the threshold for modified files manually: `git show 
8265271 -M10`

psafont:
I feel this is big enough of a change to be merged, other updates can come down the line.

psafont:
The 2 patches that were removed were fixed upstream:
* https://github.com/ocaml/num/pull/6
* https://github.com/janestreet/core/issues/119#issuecomment-443809585

PR #407, opened 2019-08-13
Add tool to import opam files into xs-extra

psafont:
It also adds the url fields as necessary, reducing handywork needed to maintain the duplicated opam files.

### DIFF
--- a/tools/generate-opam-sources.sh
+++ b/tools/generate-opam-sources.sh
@@ -1,0 +1,71 @@
+#! /usr/bin/env bash
+# Emit a table of opam file and where to download it. This can be used
+# to update Opam files to the version in a code repository
+
+EXTRA="packages/xs-extra"
+GH="https://raw.githubusercontent.com"
+XAPI="xapi-project"
+XSER="xenserver"
+
+MAP="\
+$EXTRA/nbd-tool.master/opam                  $GH/$XAPI/nbd/master/nbd-tool.opam
+$EXTRA/rrd2csv.master/opam                   $GH/$XSER/rrd2csv/master/rrd2csv.opam
+$EXTRA/xen-api-client-lwt.master/opam        $GH/$XAPI/xen-api-client/master/xen-api-client-lwt.opam
+$EXTRA/xen-api-client.master/opam            $GH/$XAPI/xen-api-client/master/xen-api-client.opam
+$EXTRA/xen-api-client-async.master/opam      $GH/$XAPI/xen-api-client/master/xen-api-client-async.opam
+$EXTRA/xapi-database.master/opam             $GH/$XAPI/xen-api/master/xapi-database.opam
+$EXTRA/xapi-client.master/opam               $GH/$XAPI/xen-api/master/xapi-client.opam
+$EXTRA/xapi-consts.master/opam               $GH/$XAPI/xen-api/master/xapi-consts.opam
+$EXTRA/xapi-types.master/opam                $GH/$XAPI/xen-api/master/xapi-types.opam
+$EXTRA/xapi.master/opam                      $GH/$XAPI/xen-api/master/xapi.opam
+$EXTRA/xapi-cli-protocol.master/opam         $GH/$XAPI/xen-api/master/xapi-cli-protocol.opam
+$EXTRA/xapi-datamodel.master/opam            $GH/$XAPI/xen-api/master/xapi-datamodel.opam
+$EXTRA/xe.master/opam                        $GH/$XAPI/xen-api/master/xe.opam
+$EXTRA/xapi-xenopsd-xc.master/opam           $GH/$XAPI/xenopsd/master/xapi-xenopsd-xc.opam
+$EXTRA/xapi-xenopsd.master/opam              $GH/$XAPI/xenopsd/master/xapi-xenopsd.opam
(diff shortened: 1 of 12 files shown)


tools/generate-opam-sources.sh
```diff
@@ -0,0 +1,80 @@
...
+xapi-rrd-transport-utils  $XAPI/rrd-transport
+xapi-rrd-transport        $XAPI/rrd-transport
+xapi-rrdd                 $XAPI/xcp-rrdd
+rrd-transport             $XAPI/rrd-transport
+xenops                    $XAPI/xenops
+xapi-xenops               $XAPI/xenops
+xapi-xenops-utils         $XAPI/xenops
+wsproxy                   $XAPI/wsproxy
+xapi-idl                  $XAPI/xcp-idl
+xapi-tapctl               $XAPI/tapctl"
```
edwintorok: Can you use `opam info -f url.src: ${package}` to get the URLs directly from opam? e.g. `opam info -f url.src: xapi-datamodel `gives `"https://github.com/xapi-project/xen-api/archive/master/master.tar.gz"`

psafont: Hmm, that would need to get opam to be configured properly. The way the script is set up makes it independent from opam.

In any case we still need that url chunk to curl the opam file from raw.githubusercontent.com

tools/generate-opam-sources.sh
```diff
@@ -0,0 +1,80 @@
...
+echo "$MAP" | while read -r opam repo; do
+  opam_file="$EXTRA/$opam.master/opam"
+  url_source="\
+url {
+  src: \"https://github.com/$repo/archive/master.tar.gz\"
+}"
+
+  curl -L https://raw.githubusercontent.com/"$repo"/master/"$opam".opam > "$opam_file"
+  # do not add the field if it's already in the opam file
+  if ! grep -F "url {" "$opam_file" > /dev/null; then
```
lindig: I believe `grep -q -F ` would not require the redirection.

```
      -q, --quiet, --silent
              Quiet;  do  not  write anything to standard output.  Exit immediately with zero status if any
              match is found, even if an error was detected.  Also see the -s or --no-messages option.
```

tools/generate-opam-sources.sh
edwintorok: If the script can run without opam that is a useful property to preserve.
What if you fetch the github URL from the existing opam file instead? i.e. grep for either the dev-repo or the source url and use sed to find the parts you need and then put it under raw.githubusercontent. Then we don't need to maintain this list of mappings (in case we get more subpackages in the future, in particular this doesn't include the varstored-guard opam file for example)

tools/generate-opam-sources.sh
psafont: I'll use `-q`, I got a warning for trying to use fgrep, though:
```
fgrep is non-standard and deprecated. Use grep -F instead.
```

tools/generate-opam-sources.sh
lindig: There is an ocaml tool to extract fields from Opam files: 

* https://opam.ocaml.org/packages/opam-ed/

Maybe we should add it to xs-extra such that we have it available?

psafont: I don't like maintaining this either.

I can try using a list of "watched" repositories, the script would download all the opam files it finds on those, create directories and files and add the `url.src` to them as needed.

This would mean the source of how xs-extra is populated remains inside xs-opam, while on the repos we can maintain the files as needed.

Ideally I'd also like some kind of CI that auto-updates xs-opam when detecting a difference after running this script.

psafont: I think we can leave this aspect of the PR as-is and open another PR to look for missing opam files in watched repositories.

psafont:
I'm updating the opam files on some out-of-date repos. After that I'll add a commit with the refresh coming from the script.

tools/generate-opam-sources.sh
```diff
@@ -0,0 +1,81 @@
...
+uuid                      $XAPI/xen-api-libs-transitional
+http-svr                  $XAPI/xen-api-libs-transitional
+sexpr                     $XAPI/xen-api-libs-transitional
+pciutil                   $XAPI/xen-api-libs-transitional
+stunnel                   $XAPI/xen-api-libs-transitional
+gzip                      $XAPI/xen-api-libs-transitional
+sha1                      $XAPI/xen-api-libs-transitional
+xml-light2                $XAPI/xen-api-libs-transitional
+zstd                      $XAPI/xen-api-libs-transitional
+rrdd-plugin               $XAPI/ocaml-rrdd-plugin
```
edwintorok: A few missing packages here:
```
rrddump
xapi-compression
xenctrl
```

edwintorok: Also the packages from `packages/xs` but these change more rarely.

psafont: I think those are the only ones in xs-extra that are missing from the map, if we don't count the weird `xcp.transition`.

psafont: The scripts only tracks packages from `packages/xs-extra`, it's not meant to keep track of versioned opam files like the ones in `packages/xs`

PR #409, opened 2019-08-23
Add 4.08.1 for testing

edwintorok:
Thanks to @psafont update of opam packages there is now very little preventing us from starting to test 4.08 (just ppx_tools which requires a different version).

### DIFF
--- a/tools/travis.sh
+++ b/tools/travis.sh
@@ -16,12 +16,16 @@ else
     INSTALL="opam install xs-toolstack"
 fi
 
+if [ "${OCAML_VERSION}" = "" ]; then
+    OCAML_VERSION="4.07"
+fi
+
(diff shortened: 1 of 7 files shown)


tools/travis.sh
```diff
@@ -16,12 +16,16 @@ else
...
+    OCAML_VERSION="4.07"
+fi
+
 docker run --rm -iv $PWD:/mnt $IMG <<EOF
 set -e
 sudo apt-get update
 opam repo remove --all default
 opam repo add xs-opam --all-switches file:///mnt
-opam switch 4.07
+opam switch ${OCAML_VERSION}
```
lindig: Could be written as `${OCAML_VERSION:-4.07}` and dropping the `if` above.

lindig:
Any reason to keep this as draft? I'd merge it.

PR #411, opened 2019-08-28
Update upstream packages

psafont:
The are some updates to dune, bugfixes, some packages have breaking changes but they're not direct dependencies of the toolstack.

The weirdest thing is core_kernel, which was updated but the folder did not reflect that.


packages/upstream/cohttp-async.2.3.0/opam
```diff
@@ -24,21 +24,23 @@ doc: "https://mirage.github.io/ocaml-cohttp/"
...
   "ocaml" {>= "4.04.1"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {>= "1.1.0"}
   "async_kernel" {>= "v0.11.0"}
   "async_unix" {>= "v0.11.0"}
   "async_extra" {>= "v0.11.0"}
   "base" {>= "v0.11.0"}
   "core" {with-test}
-  "cohttp"
+  "cohttp" {=version}
```
lindig: Is this correct? I assume this enforces that co-located package are all on the same version.

packages/upstream/cohttp-lwt-unix.2.3.0/opam
```diff
@@ -26,13 +26,13 @@ doc: "https://mirage.github.io/ocaml-cohttp/"
...
   "ocaml" {>= "4.04.1"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {>= "1.1.0"}
   "conduit-lwt-unix" {>= "1.0.3"}
   "cmdliner"
   "magic-mime"
   "logs"
   "fmt" {>= "0.8.2"}
-  "cohttp-lwt"
+  "cohttp-lwt" {=version}
```
lindig: Again - this seem unusual.

packages/upstream/cohttp-async.2.3.0/opam
mseri: Yes, we are starting using this in some packages that have very strong interdependency upstream

PR #414, opened 2019-09-26
Bump upstream packages.

psafont:
Most of them are minor updates.
Biggest change that needs to be tested is conduit and ssl: https://github.com/mirage/ocaml-conduit/pull/291/files

Only ounit is not kept up-to-date with upstream, as it may have some more updates incoming.


psafont:
I've removed the breaking changes from conduit, let me run another build with these changes to rule out unpleasant surprises

psafont:
Xenserver build is green, I think this PR is ready to be merged

PR #416, opened 2019-10-18
CP-32138: Add logs-syslog and dependencies

psafont:
These packages facilitate https://github.com/xapi-project/message-switch/issues/39.

I still need to test my changes for message switch, but I think it's pretty certain that logs-syslog will be used in any case.

The syslog library could potentially replace our custom bindings for syslog in xcp-idl to avoid repetition (lwt_log had its own implementation as well)


edwintorok:
I see, so this talks directly to `/dev/log` instead of having a C binding to syslog, probably better for concurrency because we can do non-blocking IO on it.

PR #417, opened 2019-10-24
Add crowbar

psafont:
Also updated ocp-indent checksums, somehow they have changed and broke installation.

### DIFF
--- a/packages/upstream-extra/ocp-indent.1.8.1/opam
+++ b/packages/upstream-extra/ocp-indent.1.8.1/opam
@@ -51,7 +51,7 @@ post-messages: [
 url {
   src: "https://github.com/OCamlPro/ocp-indent/archive/1.8.1.tar.gz"
   checksum: [
-    "md5=a62d50b18ded0645b3350fdc25910e87"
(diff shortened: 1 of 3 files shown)


lindig:
This adds crowbar to `xs-extra`, which is not going into the RPM which we are building. It has dependencies outside `xs-opam` but since we are not asking to build it, I don't think it matters.

packages/upstream-extra/ocp-indent.1.8.1/opam
```diff
@@ -51,7 +51,7 @@ post-messages: [
 url {
   src: "https://github.com/OCamlPro/ocp-indent/archive/1.8.1.tar.gz"
   checksum: [
-    "md5=a62d50b18ded0645b3350fdc25910e87"
-    "sha512=9257017fd5013687e453416107a763d807fd1b27a02998b6ca99b8e554747c5f2df78c2d22f0d57594e7eea76854ff6b8186ee530657ee1b33d2443630afef2e"
+    "md5=70db6649a8c08a682ad63730c9752e31"
```
lindig: Identical to upstream http://opam.ocaml.org/packages/ocp-indent/

psafont:
I've added afl-persistent so travis can install crowbar.

PR #418, opened 2019-10-29
New release for xcp-rrd

psafont:
We forgot about this one this morning!


packages/xs/xapi-rrd.1.7.0/opam
```diff
@@ -9,26 +9,28 @@ tags: [
...
+  ["dune" "runtest" "-p" name]
 ]
 depends: [
   "ocaml"
   "dune" {build & >= "1.4.0"}
   "base-bigarray"
   "base-unix"
-  "ppx_deriving_rpc"
-  "rpclib"
+  "rpc" {>= "1.9.51" & < "5.0.0"} | ("rpclib" & "ppx_deriving_rpc")
```
edwintorok: I think we could've dropped `rpc` from here completely, but lets leave it as is now.

psafont: We did, I need to change the opam file in xcp-rrd

PR #420, opened 2019-11-04
Add fmt >= 0.8.8 to xapi-xenopsd

lindig:
Xenopsd depends on Dump.record from fmt which is not present in fmt
0.8.6 at least.


psafont:
Feature was added in 0.8.7.

What situation caused fmt to be 0.8.6? fmt.0.8.8 should be the only version available in xs-opam since 4387e4dc9cff2f187be1844bc6dad5e443e39105

edwintorok:
I think it is 0.8.6 if you used an older xs-opam and didn't update.
Even though we only carry one version in xs-opam already installed packages are not automatically updated when a package version disappears from it.

PR #421, opened 2019-11-29
Use OCaml 4.08 by default, 4.09 speculatively

lindig:
* Use OCaml 4.08 by default
* Add the OCaml 4.09 compiler
* Compile with OCaml 4.09 but allow it to fail
* Compiling with OCaml 4.09 currently fails because of version constraints in upstream packages


psafont:
The ocaml version in https://github.com/xapi-project/xs-opam/blob/master/tools/xs-opam-ci.env should be changed as well to update the packages' CI

PR #425, opened 2020-01-15
REQ-453: add function to x509 lib to retrieve signature algorithm

psafont:
This is needed to enforce SHA256 as the hash signature algorithm as the current released version does not expose the signature algorithm of a certificate in any way.

There is a will to add a function to disallow signature algorithms in x509, so we might be able to use that when the time comes, see https://github.com/mirleft/ocaml-x509/issues/123

### DIFF
--- a/packages/upstream/x509.0.8.1+1/files/0001-signature-algorithm.patch
+++ b/packages/upstream/x509.0.8.1+1/files/0001-signature-algorithm.patch
@@ -1,0 +1,26 @@
+diff --git a/lib/certificate.ml b/lib/certificate.ml
+index 6297dd3..5878d73 100644
+--- a/lib/certificate.ml
++++ b/lib/certificate.ml
+@@ -196,6 +196,8 @@ let validity { asn ; _ } = asn.tbs_cert.validity
+ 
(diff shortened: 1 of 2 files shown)


packages/upstream/x509.0.8.1+1/files/0001-signature-algorithm.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/lib/x509.mli b/lib/x509.mli
+index d94f924..30bc0cb 100644
+--- a/lib/x509.mli
++++ b/lib/x509.mli
+@@ -427,6 +427,8 @@ module Certificate : sig
+ 
+   (** [extensions certificate] is the extension map of [certificate]. *)
+   val extensions : t -> Extension.t
++
++  val signature_algorithm : t -> ([> `ECDSA | `RSA ] * [> `MD5 | `SHA1 | `SHA224 | `SHA256 | `SHA384 | `SHA512 ]) option
```
edwintorok: This should perhaps have a type defined, but since this is a just temporary change until upstream updates a smaller change like here is probably better.

PR #430, opened 2020-02-04
CP-29837: use odig to build docs for all packages

psafont:
This allows the generation of html documentation for all the ocaml packages used in the toolstack, still need to think how to store/show/host the html :)


lindig:
Is this the correct path and package name? `packages/upstream-extra/b0.0.0.0/opam`

psafont:
yeah, it's the correct name with the correct version ¯\_(ツ)_/¯

PR #431, opened 2020-02-06
maintenance: update travis build instructions

lippirk:
Signed-off-by: lippirk <ben.anson@citrix.com>


README.md
```diff
@@ -66,5 +65,6 @@ script: bash -ex .travis-docker.sh
 env:
   global:
     - PINS="EXAMPLE:."
+  jobs:
```
psafont: I don't think there's a need to add jobs when there's only one.
It might be more useful to pin 2 packages in global, then have two jobs, one per package

PR #433, opened 2020-02-12
CP-32663: Update x509 to 0.9.0

psafont:
This introduces breaking changes and needs an updated xen-api: https://github.com/xapi-project/xen-api/pull/4032


lindig:
Before we can merge xen-api.git, we need to merge this PR and make a new release to avoid breaking the build outside of Travis.

lindig:
Do we still need this? If so, I would merge it tomorrow and release xs-opam.

psafont:
This will eventually go in for the certificate work, might as well do it early.

PR #440, opened 2020-03-13
pin packages for quebec

lippirk:


psafont:
This might take a while to review... the branching point for both the origin and destination branches look good.

packages/xs-extra/xapi-rrd-transport.1.16.0/opam
```diff
@@ -6,12 +6,12 @@ dev-repo: "git+https://github.com/xapi-project/rrd-transport.git"
 bug-reports: "https://github.com/xapi-project/rrd-transport"
 depends: [
   "ocaml"
-  "rrd-transport" {>= "2.0.0"}
+  "rrd-transport" {>= "1.16.0"}
```
psafont: better to use `{=version}`

packages/xs-extra/xapi-rrd-transport-utils.1.16.0/opam
```diff
@@ -9,12 +9,12 @@ depends: [
   "ocaml"
   "dune" {build & >= "1.0+beta10"}
   "cmdliner"
-  "xapi-rrd-transport" {>= "2.0.0"}
+  "xapi-rrd-transport" {>= "0.9.0"}
```
psafont: better to use `{=version}`

packages/xs-extra/xenctrl.master/opam
```diff
@@ -22,7 +22,8 @@ depexts: [
   ["xen-dev"] {os-distribution = "alpine"}
 ]
 
-dev-repo: "git+https://github.com/lindig/xenctrl.git"
+dev-repo: "git+https://github.com/xapi-project/xenctrl.git"
 url {
-  src: "https://github.com/lindig/xenctrl/archive/master.tar.gz"
+  src: "https://github.com/xapi-project/xenctrl/archive/master.tar.gz"
+  checksum: "sha256=06d21d098f2a597b1ebeb80f14150581c8a7a937d95227d40f24fbf412e0c650"
```
psafont: This will break once the commit changes, please remove the checksum.

lippirk:
i removed the versions on the rrd packages - it doesn't really make sense to apply version constraints, since all the packages are pinned anyway

psafont:
Why is `rrdd-plugins` back to master?

lippirk:
> Why is `xs-extra-dummy/rrdd-plugins` back to master?

it's just a dummy package so doesn't really have a version

psafont:
It's mistagged as a dummy plugin then (see master), please move it to xs-extra with the the version 1.9.0

lippirk:
> It's mistagged as a dummy plugin then (see master), please move it to xs-extra with the the version 1.9.0

i think you mean 1.6.7?

tools/Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM        ocaml/opam2:debian-9-ocaml-4.07
+FROM        ocaml/opam2:debian-10-ocaml-4.07
 MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
 
 COPY --chown=opam:opam . /tmp/xs-opam
 
 RUN sudo apt-get update
+RUN sudo apt-get install -qq -yy hwdata libdlm-dev libffi-dev libgmp-dev \
```
lindig: I am not sure these are strictly necessary because the `opam depext -y xs-toolstack` should install them. We once had problems with that, though.

lippirk: ah you're right - without this line, building the image does work, it just hangs for a while and gives a message which makes it seem like it's going to fail

PR #442, opened 2020-03-22
Print more log lines on build failures

edwintorok:
This should hopefully show us what the build failure is


edwintorok:
Doesn't seem to work, still prints just very few lines

edwintorok:
PR to fix build here: https://github.com/xapi-project/xen-api/pull/4062

lindig:
Is this too much? But I agree in principle that more lines are useful.

edwintorok:
We might have to reduce the number if we find out its too much and we hit the Travis log file size limit. However 80 chars * 10000 lines is only ~800 KB

PR #443, opened 2020-03-24
merge rrddump with xcp-rrdd

lippirk:


lindig:
I'm re-running the Travis build now that we merged rrddump into xcp-rrdd

PR #444, opened 2020-03-24
Maintenance: lint and format opam files

lindig:
Reformat opam files using "opam lint -normalise" in the xs/and xs-extra/
hierarchy.


lippirk:
it might be worth posting the command you used to do this so we can double check it makes sense

lindig:
```
#! /bin/bash

opam lint --normalise $1 > $1.tmp && mv $1.tmp $1
```

```
$ find xs xs-extra xs-extra-dummy -name opam -type f -exec ../norm.sh '{}' \;

**Comment by psafont** [comment]
We need to be careful when importing changes using ` xs-opam/tools/generate-opam-sources.sh` as the changes in here need to be applied upstream as well.

File: packages/xs-extra/xenops-cli.master/opam [hunk 1]
```diff
@@ -1,27 +1,28 @@
...
-"""
+build: [
+  ["dune" "build" "-p" name]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+dev-repo: "git+https://github.com/xapi-project/xenops-cli.git"
 url {
-  src: "https://github.com/xapi-project/xenops-cli/archive/master/master.tar.gz"
+  src:
+    "https://github.com/xapi-project/xenops-cli/archive/master/master.tar.gz"
```
psafont: This is out of date, it should point to xenopsd now https://github.com/xapi-project/xenopsd/pull/692

psafont: https://github.com/xapi-project/xs-opam/blob/master/tools/generate-opam-sources.sh#L68needs to be updated as well to reflect this

lindig:
When the general flow of information is from repos to here, we should not take this but rather reformat the opam files in the repos.

psafont:
we have an automated way to import the files from the repo, could there be an automated way to push it to the repos as PRs?

lindig:
I think we should close this and reformat the opam files in the repos. I don't think there is a good way to automate this but maybe we can just do it when we work on a repo given how simple it is.

lindig:
Closing this as discussed above.

PR #449, opened 2020-04-01
REQ-811

lippirk:


lippirk:
preemptive PR (nbd not merged yet) to pin nbd in xs-opam

psafont:
Can it be moved to `packages/xs/`? By convention all packages in `/packages-xs-extra/` use `master`.

For people worried about the transition from master to a version, this can alleviate problem on the dependent packages when developing:

> `pin-depends: [ [ <package> <URL> ] ... ]`: this field has no effect on the package repository, but is useful for in-source specification of development packages. When source-pinning the package, either through `opam pin` or `opam install <DIR>`, opam will prompt to pin every specified `<package>` to the associated `<URL>`. There are two important limitations:
> - If you want the pinned package be a dependency you need to add its `<pkgname>` to depends: field.
> - `pin-depends`: are NOT transitive, that is, pin-depends: of packages getting pinned through pin-depends: are ignored
> - They won't get updated on opam update, the users will need to use opam pin or opam install|upgrade DIR again to get the new pins if the field has changed. Even then, this won't unpin any packages that would have been removed from `pin-depends:`.

PR #462, opened 2020-05-14
CA-334811 set dummy XAPI_VERSION var

lippirk:
sibling of https://github.com/xapi-project/xen-api/pull/4100


lippirk:
no longer required

PR #463, opened 2020-05-29
add mtime

edwintorok:
xapi-idl's opam file got updated, but xs-opam didn't causing Travis to fail elsewhere.


psafont:
It's not the only file that's out-of-sync.
Could you run `tools/generate-opam-sources.sh`, ignore the changes in `packages/xs-extra/xenctrl.master/opam` and commit the rest?

edwintorok:
That removes `fpath` from `xe.master`, but I think we're going to need it, so I added it back.

PR #472, opened 2020-06-24
CI: ensure env vars are picked correctly on github actions and PRs trigger builds

psafont:
Github changed how the environment variables from the matrix can be picked up, this corrects it.

fail-fast is disabled to be able to see all steps that fail and continue-on-error is enabled for the cases that are expected to fail. This masks the errors but not masking them makes github send notifications and emails reporting failures. There doesn't seem to be a fix yet for this behaviour.

The recursive job is aborted in travis due to timeout, with github this doesn't seem to be the case, it runs for a _long_ time. I'd consider fixing it or disabling it altogether.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
         env: [OPAMWITHTEST: false,
               OPAMWITHTEST: true,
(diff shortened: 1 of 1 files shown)


psafont:
The last commit cannot be tested beforehand as the branch I pushed doesn't match the patterns defined in the branches.

edwintorok:
Does the recursive one ever produce a result?

psafont:
All the ones I've seen I cancelled them, even after running 5 hours

PR #473, opened 2020-06-25
CI: use valid names whe nexporting env vars

psafont:
this was breaking travis builds in toolstack components


edwintorok:
Don't you need to update the github CI configuration somewhere too to take into consideration the new env var?

edwintorok:
Also I was expecting to see a Github CI action to verify this PR, do we need to enable some integration for that to happen?

psafont:
> Don't you need to update the github CI configuration somewhere too to take into consideration the new env var?

I'll update xen-api's PR to expect the different name for the key

> Also I was expecting to see a Github CI action to verify this PR, do we need to enable some integration for that to happen?

We need to merge the other open PR for that :)

PR #475, opened 2020-06-29
Travis: use debian-10-ocaml-4.08

lindig:
We fall back to debian-10 to avoid a problem with the current
debian-unstable images.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,10 +1,11 @@
-FROM        ocaml/opam2:debian-10-ocaml-4.08
+FROM ocaml/opam2:debian-10-ocaml-4.08
 MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
 
 COPY . /tmp/xs-opam
 
 RUN sudo apt-get update
 RUN opam repo remove --all default \
- && opam repo add xs-opam file:///tmp/xs-opam \
+ && opam repo add xs-opam --all-switches file:///tmp/xs-opam \
+ && opam switch 4.08 \
  && opam depext -y xs-toolstack \
  && opam install xs-toolstack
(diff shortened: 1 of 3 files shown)


tools/Dockerfile
```diff
@@ -1,10 +1,11 @@
...
+FROM ocaml/opam2:debian-10-ocaml-4.08
 MAINTAINER  Christian Lindig <christian.lindig@citrix.com>
 
 COPY . /tmp/xs-opam
 
 RUN sudo apt-get update
 RUN opam repo remove --all default \
- && opam repo add xs-opam file:///tmp/xs-opam \
+ && opam repo add xs-opam --all-switches file:///tmp/xs-opam \
+ && opam switch 4.08 \
```
lippirk, line 9: 4.08.1?

tools/xs-opam-ci.env
```diff
@@ -1,6 +1,6 @@
 export OCAML_VERSION="4.08"
 export OCAML_VERSION_FULL="4.08.1"
-export DISTRO="debian-unstable"
+export DISTRO="ocaml/opam2:debian-10-ocaml-4.08"
```
psafont, line 3: The first part of the image name is filled by the script, only the distro should be exported:
```
from=${hub_user}/opam2:${DISTRO} ;;
```

tools/Dockerfile
lindig: When I tested it locally, it worked. Not sure how significant this is.

lindig:
Is this not triggering a Travis run?

tools/Dockerfile
lippirk: ignore this then :+1:

psafont:
> Is this not triggering a Travis run?

It is: https://travis-ci.org/github/xapi-project/xs-opam/builds/703182059

I'm more surprised at github ci not triggering

PR #476, opened 2020-06-29
github actions: enable builds on pull requests

psafont:
The indentation was incorrect, blocking builds.


psafont:
I've cancelled the --recursive job after more than an hour: it sometimes fails like this, the test it not reliable.

I could set a shorter timeout of around 40 minutes if it's needed.

PR #477, opened 2020-06-30
ci: remove job for testing all upstream packages

psafont:
It never succeeded and there's little chance this can change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ jobs:
   include:
     - env: OPAMWITHTEST=false
     - env: OPAMWITHTEST=true
-    - env: OPAMWITHTEST=true RECURSIVE=--recursive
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,8 +10,6 @@ jobs:
   include:
     - env: OPAMWITHTEST=false
```
lippirk, line 11: why do we bother with `OPAMWITHTEST=false`?

psafont: I don't think it makes sense either, I've updated the PR

psafont:
Travis job failed because of the pci identifiers changed, it was fixed by https://github.com/xapi-project/xen-api/pull/4150 but I cannot rerun the tests. the job for Github 4.08 builds and tests the packages using the same containers.

PR #481, opened 2020-07-06
maintenance: xenctrl.dummy depends on xen libraries

lippirk:
This should fix the travis failures for stockholm


psafont:
the github ci complains about the `dummy` version not matching  metadata `0` is this something that is needed?

lippirk:
> the github ci complains about the `dummy` version not matching metadata `0` is this something that is needed?

let's try removing it

psafont:
> edit: this works from travis' point of view. How can I actually see the output of github ci?

We would need to merge https://github.com/xapi-project/xs-opam/pull/480 first, the failure was on a xen-api PR.

PR #484, opened 2020-07-09
ci: add github workflow for lcm

psafont:
To trigger the building and testing on a particular commit the workflow file must be on its predecessor. This applies for prs, commits and tags.

Unfortunately scheduled runs for this branch have to be triggered by a workflow on the main branch this means we must manage two workflows for lcm, one here, one on the main branch.
I'll repurpose the existing lcm workflow in `master` to schedule runs for this branch.

### DIFF
--- a/.github/workflows/lcm.yml
+++ b/.github/workflows/lcm.yml
@@ -1,0 +1,22 @@
+name: Build and test (Stockholm)
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


psafont:
Both agents are having problems downloading menhir, I've had to rerun the github job a couple times to get beyond that error.

PR #488, opened 2020-07-10
xs-extra: update opam metadata from repositories

psafont:
this is done by running tools/generate-opam-sources.sh


psafont:
green status depends on https://github.com/xapi-project/xs-opam/pull/487

PR #489, opened 2020-07-13
CP-33121: Remove obsolete stdext modules

psafont:
Finally remove quite a few stdext packages that have been made redundant with the third-party libraries and standard libraries of recent ocaml versions.

Remaining libraries still have some specialized, useful methods not easily available on stdlib. I'd like to remove the usage of them with 'open' to make them be specialized libraries for those methods instead of simply being an overlay on top of stlib. This is egregious in xen-api's use of mutex, list and string modules.

The current way creates uncomfortable positions: the new stlibs expose more functions than these libraries; the new stlibs expose functions implemented in these libraries and it gets blocked.


edwintorok:
Looks good, we probably want actual release numbers there instead of 'nomonadic'.

psafont:
Yes, this should not get merged as-is, I'm tagging a new release of stdext to be able to integrate it into xs-opam.

PR #490, opened 2020-07-16
maintenance: rename xenctrl.dummy to xenctrl.0~dummy

lippirk:
With this change, the xen-api stockholm branch will now
compile on travis.

opam prefers branch names over version numbers, meaning
that xenctrl.1.1.0 is not chosen when you try to compile
a package depending on xenctrl - instead opam will pick
xenctrl.dummy. However xenctrl dummy does not compile,
leading to failing builds.

However if we use 0~dummy instead of dummy, opam will
correctly choose 1.1.0.


lippirk:
With some modification to `xs-opam-env` in `xs-opam` and `.travis-xs-opam.sh` in `xen-api`, I got a green xapi stockholm build: https://travis-ci.com/github/lippirk/xen-api/jobs/361637353

mseri:
What if instead you rename `xenctrl.dummy` to `xenctrl.0.dummy`?

lindig:
This is rooted in the way opam selects a version. On branch `master`, we have:

```
./packages/xs/xenctrl.dummy
./packages/xs-extra/xenctrl.master
```

On branch `release/stockholm/lcm` we have:

```
./packages/xs/xenctrl.dummy
./packages/xs-extra/xenctrl.1.1.0
```

Opam will select the highest version and this is:

```
: xs-opam $ ( echo dummy; echo master; echo "1.1.0" ) | sort
1.1.0
dummy
master
```

So it will select dummy over 1.10 and master over dummy.

lippirk:
I have tested a xen-api build with `xenctrl.0~dummy`, and it seemed to work :+1: I think we should consider cherry-picking this for master to avoid this problem in the future

PR #491, opened 2020-07-16
Move ocamlformat to upstream/ so we build it by default

edwintorok:
It is available in `upstream-extra`, but `opam` doesn't work inside a
pxc/mock chroot anymore because it doesn't have permission to install to
/usr (only an rpm package has).
With planex-buildenv it used to be possible to run things as root and
install additional opam packages.

Easiest is to move this to upstream/ so that xs-opam would build it and
make it available inside build containers.

That way contributors can have matching ocamlformat version to the one
that we use.

If you just try to `opam install ocamlformat` you get 0.14.2, which won't work because we want 0.14.1, so then you have to fiddle with setting up xs-opam to get it working.


lindig:
Moving it to upstream will include it into the RPMs that we build unless we do something to exclude it.

edwintorok:
Yes, it needs to be part of the rpm to be usable inside a container. But as long as we don't add it as a requirement for another package it won't be part of the ISO. It will increase the build times slightly, at least until we move to Koji.
I'm not sure what the best approach here should be. Having a container that you can use for development might be useful for people immediately outside of toolstack that want to contribute (e.g. Igor contributing to xenopsd).

I used to have my own container for that purpose (https://github.com/edwintorok/xs-opam-dev/), but having 2 containers might be a bit inconvenient.

lindig:
Any opam package in xs/ and upstream/ is packaged into one RPM xs-opam-repo which is used for building other RPMs. So in that sense it is part of an RPM. But this RPM is not shipped to customers and not installed on a customer's system.

PR #493, opened 2020-07-20
maintenance: replace xenops-cli with xapi-xenopsd-cli

lippirk:
xenops-cli was replaced by xapi-xenopsd-cli when
we merged xenops.git and xenops-cli.git into
xenopsd.git last year.


lippirk:
@robhoes I believe we should archive xenops-cli and xenops on xapi-project

robhoes:
@lippirk Thanks, I missed this last year. It's a pity that these opam files are duplicated, because it is quite error prone. Regarding archiving those repos, I think that I didn't do that, because we might still need them for LCM.

PR #495, opened 2020-07-21
Add licenses target to report them

lindig:
Add Makefile target "licenses" to report the licenses of all toolstack
packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL   = /bin/bash
 VERSION = $(shell git describe --abbrev=0 --tags)
 NAME 	= xs-opam-repo-$(VERSION)
 
-.PHONY: all archive clean
+.PHONY: all archive clean licenses
 
 all:
 	docker build -f tools/Dockerfile -t xenserver/xs-opam:$(VERSION) .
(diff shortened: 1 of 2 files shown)


psafont:
The command is a bit awkward to run since it implicitly needs an active switch with the xs-opam repo only, might be worth at least documenting it in the Makefile.

It would be interesting to convert this into some sort of check in the CI on the long run so we take care of the licenses in the metadata. Right now most xapi-project packages are missing the information.

lindig:
I now see what you mean: this does not work on the data in repo but on the data that has been installed from the repo, which it not intuitive. Would it be better to put this in a script under `tools/`?

psafont:
> I now see what you mean: this does not work on the data in the repo but on the data that has been installed from the repo, which it not intuitive. Would it be better to put this in a script under `tools/`?

I'd be happy with that plus the comment explaining how to run it successfully.

Makefile
```diff
@@ -24,6 +24,11 @@ $(NAME).tar.gz:
 	mv ocaml packages
 	mv upstream-extra packages
 
+licenses:
+	opam list --required-by "xs-toolstack" --recursive --short \
+	  | xargs opam info -f name:,license: | tr -d '"' \
+	  | awk '/^name:/ { name = $$2 } /^license:/ { license = $$2; printf("%s: %s\n",name, license); }'
+
```
psafont, line 31: Did you intend to keep the command in both files?

lindig: Sorry - missed adding the updated Makefile

PR #496, opened 2020-07-22
Remove unused packages

psafont:
A few packages had to be added to upstream-extras to make the tool happy.

I updated crowbar since the version we had required old, incompatible dependencies. The new version is from May.

The packages removes are oasis and other deprecated build tools, ocveralls which is replaced by bisect_ppx, and ppxfind which used to be required by ppx_deriving_rpc.

Regarding the build tools, I haven't found any evidence of any toolstack packages using it anymore, not on spec files, not on makefiles either.

These are the repositories I have checked-out that still use it:
```
fd setup.ml
libvhd/setup.ml
luna-rossa/setup.ml
ocaml-blktap3-stats/setup.ml
ocaml-pci-db/setup.ml
ocaml-xen-lowlevel-libs/setup.ml
ocaml-xenstore/setup.ml
ocaml-xenstore-old/setup.ml
ocaml-xenstore-xen/unix/setup.ml
ocaml-xenstore-xen/xen/setup.ml
oclock/setup.ml
quicktest/setup.ml
xapi-libvirt-storage/setup.ml
```


psafont:
`ocaml-xen-lowlevel-libs` is still an active project, as it keeps track of code ocaml code in xen.git

Depending on how developers environment is set up it might impact developers like you.

The rest of repositories have not been updated since at least 2016, so I doubt they are really being used.

packages/xs-extra-dummy/upstream-extra-dummy.master/opam
```diff
@@ -5,13 +5,13 @@ authors:        "Christian Lindig <christian.lindig@citrix.com>"
...
+  "odig"
   "travis-opam"
-  "merlin"
-  "xenctrl" { = "dummy"}
-  "ocveralls"
   "utop"
+  "xenctrl" { = "dummy"}
 
   # TODO: not used but we may want to keep
   "opasswd"
```
lippirk, line 17: could we remove this too?

lindig: I think we could remove it as well. But given that @psafont is not here, I'd prefer to address is another time.

psafont: @edwintorok added it a couple of years ago, don't know for what purpose

edwintorok: I think the dummy package might be needed in our internal build system, where we want to use the system provided xenctrl (built and installed outside of opam), instead of the version that we would build here.
The opam version of xenctrl is only used in the CI, and for local builds, not for the production builds.

lippirk: I was talking about opasswd rather than xenctrl

edwintorok: Ah, that was there when we were testing @lindig's unixpwd implementation vs upstream `opasswd`. `unixpwd` is now inside xapi, so this can be dropped, I'll send a PR.

PR #498, opened 2020-07-24
Update xs-extra packages and package ocamlformat's dependencies with it

psafont:
Ocamlformat was moved so its source code was packaged when `make archive` was run, but not some of its dependencies. 

This change fixes this and updates all packages in upstream-extra to the latest version.


psafont:
Move tyxml now, which is needed by odoc.

PR #505, opened 2020-08-14
Add xs/polly.0.2.0

lindig:
Polly is split out of xenopsd to make it available outside.

### DIFF
--- a/packages/xs/polly.0.2.0/opam
+++ b/packages/xs/polly.0.2.0/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+name: "polly"
+synopsis: "Bindings for the Linux epoll system call"
+description: """
+Bindings for the Linux epoll system call. The binding avoids
+  most allocation in the event loop by iterating over all file
+  descriptios that are reported as ready ."""
+maintainer: "Christian Lindig <christian.lindig@citrix.com>"
+authors: "Christian Lindig <christian.lindig@citrix.com>"
+license: "MIT"
+homepage: "https://github.com/lindig/polly"
+doc: "https://github.com/lindig/polly"
+bug-reports: "https://github.com/lindig/polly/issues"
+depends: [
+  "dune" {build}
+  "ocaml"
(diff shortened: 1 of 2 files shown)


packages/xs/polly.0.2.0/opam
```diff
@@ -0,0 +1,25 @@
...
+  "dune" {build}
+  "ocaml"
+  "cmdliner"
+  "base-unix"
+]
+build: ["dune" "build" "-p" name "-j" jobs "@install"]
+dev-repo: "git://github.com/lindig/polly.git"
+url {
+  src: "https://github.com/lindig/polly/archive/0.2.0.tar.gz"
+  checksum: "sha256=5513d2be1e286cb1ca67160c1903bad4e50bafcad8eeaeea9522f13675f09eecA"
```
psafont, line 24:
```suggestion
  checksum: "sha256=5513d2be1e286cb1ca67160c1903bad4e50bafcad8eeaeea9522f13675f09eec"
```

packages/xs/polly.0.2.0/opam
```diff
@@ -0,0 +1,25 @@
...
+name: "polly"
+synopsis: "Bindings for the Linux epoll system call"
+description: """
+Bindings for the Linux epoll system call. The binding avoids
+  most allocation in the event loop by iterating over all file
+  descriptios that are reported as ready ."""
+maintainer: "Christian Lindig <christian.lindig@citrix.com>"
+authors: "Christian Lindig <christian.lindig@citrix.com>"
+license: "MIT"
+homepage: "https://github.com/lindig/polly"
```
edwintorok, line 11: Should we move this inside the xapi project?

edwintorok: Or we could leave it there for now, if we intend to merge the repositories it'll end up as a subtree of the merged toolstack repo anyway.

lindig: Yes - but let's make this a separate step.

PR #511, opened 2020-09-25
Update dependencies to be compatible with ocaml 4.10

psafont:
In order to get a successful CI, the repositories in xs-extra that were pinned in the branch need to be updated as well. I'm opening those Pull requests as well today.

This branch comes from the feature/4.10 one, which can be discarded once we merge this one.


lippirk (approved):

edwintorok (approved):
Travis failure here is expected due to reverting of pins, but we can see that an earlier commit just before the revert had a green tick.

PR #513, opened 2020-10-06
fix: update cstruct-sexp to 5.1.1

psafont:
This brings it in line with the rest of cstruct packages


lindig (approved):

edwintorok (approved):
Did we find out how the xs-opam build passed with the wrong deps?

PR #514, opened 2020-10-06
CP-34675: use ocaml 4.10.1 instead of 4.10.0

psafont:


psafont:
The build fails because the container tries to look for 4.10.0 in the opam metadata but it's not found. This needs new container image that looks for the correct compiler. (this can be done using ocurrent or rebuilding `ocaml/opam2` with 4.10.1)

lindig (approved):

psafont:
Now the script compiles and installs the requested ocaml compiler version. This means it's slower when the compiler is not available in the container, but at least it doesn't fail like before.

The next step would be to switch the github actions to stop using the container. If the container is still useful outside CI I think building it only when a new version is tagged is what would make more sense, instead of building on each revision.

The downside of this is that it will take a bit more time until we notice that it broke.

psafont:
The travis failure is due to networking issues when downloading packages.

lippirk (approved):

PR #517, opened 2020-10-27
CP-34942: update dune to 2.7.1 and bisect_ppx to 2.5.0

psafont:
Features added: 
* New Dune and Merlin way of interacting: the won't need to happen using .merlin files, so per module definitions will be possible.
* Cram tests to easily test binaries with output: https://dune.readthedocs.io/en/stable/tests.html#cram-tests
* Dune doesn't crash on very long outputs anymore
* Add native ppx instrumentation (for coverage reporting): https://dune.readthedocs.io/en/stable/instrumentation.html

Not sure what the cesxp library can do, it may be worth checking whether it can replace jane street's sexp libraries to shed some dependencies.


lindig:
```
# Performing timeout tests
203
# Caught timeout exception after 4.211981 seconds
204
# 
205
# Completed timeout test
206
# 
207
# Completed exitcode tests
208
# Starting 540 tests
209
# 
210
    1   0 Fatal error: exception Forkhelpers.Subprocess_failed(2)

**Comment by lindig** [comment]
```

psafont:
Isn't `Unix.EACCES` code `2`?

This looks related to the change to get forkexec's test to run without opening `/var/xapi`:
```
$ dune exec src/fe_main.exe
Fatal error: exception Unix.Unix_error(Unix.EACCES, "mkdir", "/var/xapi")
```

psafont:
The test script works when executed directly, without dune. It seems is interfering somehow.

I've asked the dune devs and the sandboxing should not be used in this case, as it's only activated for cram tests.

psafont:
Forkexecd issue is because the test depend on stdin not being /dev/null, which dune 2.7 enforces when running executables / scripts. Fix is in https://github.com/xapi-project/forkexecd/pull/45

lindig (approved):

lippirk (approved):

PR #519, opened 2020-11-06
REQ-403: set up branch 

psafont:
Both xen-api and xen-api-libs-transitional have changes, track them with this branch.

### DIFF
--- a/packages/xs-extra/gzip.master/opam
+++ b/packages/xs-extra/gzip.master/opam
@@ -18,5 +18,5 @@ These libraries are provided for backwards compatibility only.
 No new code should use these libraries."""
 url {
   src:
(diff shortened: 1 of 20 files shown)


lippirk (approved):
The urls are confusing, because the branch is called `feature/REQ-403` and not `feature/REQ-403/master`, however they do appear to fetch the correct tarballs

psafont:
This PR is wrong, it should not be merged to master.

lindig:
This looks good but do the corresponding branches in the other repos exist already?

lindig (approved):

psafont:
Branches in xen-api and xen-api-libs-transitional have been set up.

PR #523, opened 2020-11-25
readme: replace ci instructions to use github actions

psafont:
Travis now has a restrictive limit on number of minutes per month.

No repos in xapi-project have hit it yet, but it's good practice to prepare for it.

The template given should be enough to kickstart github actions any of the repos that are yet to be migrated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,30 +89,66 @@ Packages are organised into namespaces:
 * `xs`: packages required for xs-extra
 * `xs-extra`: toolstack components - latest version
 
-## Travis
+## Continuous Integration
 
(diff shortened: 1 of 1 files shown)


edwintorok:
Can we turn this into a workflow template?
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template

psafont:
I've create the workflow template. I would like to remove the outdated travis configuration in the readme to keep it up-to-date with best practices.

edwintorok (approved):

lindig (approved):

PR #526, opened 2020-12-01
ci: detect if there are packages with multiple versions

psafont:
This avoid situations like the last merge which broke building the rpm package for xs-opam.

### DIFF
--- a/tools/opam-single-version-check.bash
+++ b/tools/opam-single-version-check.bash
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# makes sure no ordinary opam packages have more than a single version
+# the only exceptions are ocaml and xenctrl.
+# This is required because of how the rpm distributions are built: they
+# try to install all versions of all ordinary packages, which fails because
+# the versions of a single package are never compatible.
+
+packages=$(opam list --required-by "xs-toolstack" --recursive --short)
+for pkg in $packages; do
+  the_versions=$(opam info -f all-versions "$pkg")
+  versions=$(echo "$the_versions" | wc -w)
+  # the compiler is allowed to have more than one version because it's a
+  # separate rpm. xenctrl has 2 versions: dummy and master
+  if [ "$versions" -gt 1 ] && echo "$pkg" | grep -vqE '^ocaml$|^ocaml-base-compiler$|^ocaml-system$|^xenctrl$' ; then
+    >&2 echo "ERROR $pkg has multiple versions: $the_versions"
+    exit 1
+  fi
+done
+echo OK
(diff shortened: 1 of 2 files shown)


tools/opam-single-version-check.bash
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# makes sure no ordinary opam packages have more than a single version
```
lindig, line 3:
```
set -e pipefail
```

tools/opam-single-version-check.bash
```diff
@@ -0,0 +1,19 @@
...
+# the only exceptions are ocaml and xenctrl.
+# This is required because of how the rpm distributions are built: they
+# try to install all versions of all ordinary packages, which fails because
+# the versions of a single package are never compatible.
+
+for pkg in $(opam list --required-by "xs-toolstack" --recursive --short); do
+  the_versions=$(opam info -f all-versions "$pkg")
+  versions=$(echo "$the_versions" | awk -F' ' '{print NF}')
```
lindig, line 11: could be also done with `wc -w`

tools/opam-single-version-check.bash
psafont: The subshell that gather packages also needs to be moved outside of the for structure to fail gracefully.

lindig (approved):

tools/opam-single-version-check.bash
```diff
@@ -0,0 +1,22 @@
...
+# try to install all versions of all ordinary packages, which fails because
+# the versions of a single package are never compatible.
+
+packages=$(opam list --required-by "xs-toolstack" --recursive --short)
+for pkg in $packages; do
+  the_versions=$(opam info -f all-versions "$pkg")
+  versions=$(echo "$the_versions" | wc -w)
+  # the compiler is allowed to have more than one version because it's a
+  # separate rpm. xenctrl has 2 versions: dummy and master
+  if [ "$versions" -gt 1 ] && echo "$pkg" | grep -vqE '^ocaml$|^ocaml-base-compiler$|^ocaml-system$|^xenctrl$' ; then
```
psafont, line 17: Changed so to match only 4 exact patterns.

tools/opam-single-version-check.bash
```diff
@@ -0,0 +1,22 @@
...
+    echo "ERROR $pkg has multiple versions: $the_versions"
```
lippirk, line 18:
```suggestion
    >&2 echo "ERROR $pkg has multiple versions: $the_versions"
```

tools/opam-single-version-check.bash
edwintorok: How about the ppxlib? I think that needs 2 versions. One for 4.08 and one for 4.10, or does the latest version work with both?

lippirk (approved):

PR #529, opened 2020-12-10
maintenance: replace depexts with conf- packages

psafont:
The conf-* packages encapsulate depexts. This is convenient for us because header and binary dependencies are tough to define for all distributions and are package-specific.

Having a single dependency (like `conf-xen`) makes the dependencies easier to maintain. Right now we have depexts underspecified. Hopefully with this kind of change more dependencies can be added.

### DIFF
--- a/packages/xs-extra/xenctrl.master/opam
+++ b/packages/xs-extra/xenctrl.master/opam
@@ -8,19 +8,10 @@ bug-reports: "https://github.com/xapi-project/xenctrl/issues"
 depends: [
   "ocaml"
   "dune" {build}
+  "conf-xen"
   "base-unix"
 ]
 build: ["dune" "build" "-p" name "-j" jobs]
-depexts: [
-  ["m4" "libxen-dev" "libsystemd-dev"] {os-distribution = "debian"}
-  ["libxen-dev" "uuid-dev"] {os-distribution = "ubuntu"}
-  ["xen-devel" "libuuid-devel"] {os-distribution = "centos"}
-  ["xen-devel"] {os-distribution = "fedora"}
-  ["xen-devel"] {os-distribution = "rhel"}
-  ["xen-devel"] {os-distribution = "oraclelinux"}
-  ["xen-dom0-libs-devel" "xen-libs-devel"] {os = "xenserver"}
-  ["xen-dev"] {os-distribution = "alpine"}
-]
 
 dev-repo: "git+https://github.com/xapi-project/xenctrl.git"
 url {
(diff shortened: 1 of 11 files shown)


packages/xs-extra/xenctrl.master/opam
```diff
@@ -8,18 +8,13 @@ bug-reports: "https://github.com/xapi-project/xenctrl/issues"
 depends: [
   "ocaml"
   "dune" {build}
+  "conf-m4"
```
psafont, line 11: Does it really need m4?

packages/xs-extra/xenctrl.master/opam
```diff
@@ -8,18 +8,13 @@ bug-reports: "https://github.com/xapi-project/xenctrl/issues"
...
 depexts: [
-  ["m4" "libxen-dev" "libsystemd-dev"] {os-distribution = "debian"}
-  ["libxen-dev" "uuid-dev"] {os-distribution = "ubuntu"}
-  ["xen-devel" "libuuid-devel"] {os-distribution = "centos"}
-  ["xen-devel"] {os-distribution = "fedora"}
-  ["xen-devel"] {os-distribution = "rhel"}
-  ["xen-devel"] {os-distribution = "oraclelinux"}
-  ["xen-dom0-libs-devel" "xen-libs-devel"] {os = "xenserver"}
-  ["xen-dev"] {os-distribution = "alpine"}
+  ["libsystemd-dev"] {os-distribution = "debian"}
```
psafont, line 17: Does it really need the systemd headers? This seems more appropiate for the systemd bindings in the transitional libs.

I also removed the libuuid headers dependency here. They might be used elsewhere but not here.

edwintorok: xenstored needs systemd, but not xenctrl

packages/xs-extra/xenctrl.master/opam
edwintorok: No, although we're not really using this library, we're using the copy in Xen.

lippirk:
```
 #=== ERROR while compiling conf-python-2-7.1.1 ================================#
Sys_error("/home/runner/.opam/repo/default/packages/upstream/conf-python-2-7.1.1/files/test.py: No such file or directory")

#=== ERROR while compiling conf-pam.1 =========================================#
Sys_error("/home/runner/.opam/repo/default/packages/upstream/conf-pam.1/files/main.c: No such file or directory")
```

We're missing some files.. It's a shame we can't just point xs-opam to upstream conf-pam (etc.) rather than vendoring it

lippirk:
I think you're missing `main.c` for pam still

psafont:
I've tested this locally and it's enough to install xs-toolstack and all its dependencies in Fedora.

Curiously the changes have forced me to install the libnl3 library. It's a runtime dependency, not a link-time one for ocaml-netlink.

lippirk (approved):

lindig (approved):

PR #538, opened 2021-01-11
detect failed downloads in `opam admin cache`

edwintorok:
`opam admin cache` exits with code 0 even when it fails to download files. This results in failures in Jenkins.
Detect errors earlier by refusing to create the tarball if there were errors from `opam admin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ $(NAME).tar.gz:
 	mv packages/ocaml .
 	mv packages/upstream-extra .
 	env OPAMFETCH=wget opam admin cache |& tee cache.log
+	! grep ERROR cache.log
 	git archive --format=tar.gz --prefix=$(NAME)/ HEAD > $@
 	tar zxf $@
 	cd $(NAME) && ln -fs ../cache .


psafont (approved):

lippirk (approved):

lindig:
How does this work? Is `!` part of shell syntax when that line is executed or a Make mechanism?

psafont:
It's part of the sh syntax:

https://man.archlinux.org/man/core/bash/bash.1.en#Pipelines
> If the reserved  word  !   precedes  a pipeline, the exit status of that pipeline is the logical negation of the exit status as described above.  The shell waits for all commands in the pipeline to terminate before returning a value.

lindig:
Learned something new. I would have used `grep ERROR cache.log && false`.

lindig (approved):

lindig:
We should cherry pick this onto `master` as well.

lindig:
Notice that we are explicitly using bash in this Makefile (setting `SHELL`), which is not standard but required for the `|&` using in the recipe.

edwintorok:
Yeah I meant to use `&& false`, but then remembered about `!`

PR #541, opened 2021-01-19
Update xenstore_transport to 1.2.0

lindig:


lindig:
Sorry, I pushed this by accident to the wrong repo because there was typo in the opam file. Let's see that the CI works.

PR #543, opened 2021-01-21
maintenance: sync packages from default repo

psafont:
Also includes the script used to sync, which is pretty barebones.

### DIFF
--- a/packages/xs/ezxenstore.0.4.1/opam
+++ b/packages/xs/ezxenstore.0.4.1/opam
@@ -8,7 +8,7 @@ dev-repo: "git+https://github.com/xapi-project/ezxenstore.git"
 build: [[ "dune" "build" "-p" name "-j" jobs ]]
 depends: [
   "ocaml"
-  "dune" {build & >= "1.4"}
+  "dune" {>= "1.4"}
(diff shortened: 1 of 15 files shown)


packages/xs/ezxenstore.0.4.1/opam
```diff
@@ -8,7 +8,7 @@ dev-repo: "git+https://github.com/xapi-project/ezxenstore.git"
 build: [[ "dune" "build" "-p" name "-j" jobs ]]
 depends: [
   "ocaml"
-  "dune" {build & >= "1.4"}
```
lippirk, line 11: I doubt it makes any difference to us either way, but why doesn't the default repo have dune labelled as a build dependency?

lippirk (approved):

packages/xs/ezxenstore.0.4.1/opam
psafont: If dune were a build dependency, downgrading dune would not force the package to be rebuilt.

This means that one could built the library, downgrade dune, then try to compile a package that depends on the library, only to go 💥

lindig (approved):

PR #544, opened 2021-01-26
upstream-extra: update merlin and ocaml-lsp-server packages

psafont:
This prepares both for the newer versions of dune which do _not_ generate .merlin files.

The lsp server comes with some goodies:
- Support cancellation notifications when possible.
- Implement signature help request for functions
- Server LSP requests & notifications concurrently. Requests that require merlin are still serialized. 
- Code action to insert inferred module interface
- Filter keywords by context
- Add keyword completion
- Add go to declaration functionality to jump to a value's specification in a .mli file


edwintorok:
Can you add the lsp server to upstream-extra-dummy? Perhaps we could integrate it better in our dev/build environment then.

lindig (approved):

lippirk (approved):

PR #545, opened 2021-01-26
CA-350872: update mirage-crypto packages

psafont:
Previous versions checked an invalid property of RSA key parameters.

Following the cascade of updates needed in order to update to the latest version, it's clear that the size ballooned too much for the fix and it's worth adding the patches to the lcm version.

The 3 packages come from the same source, this is why I decided to apply the patches to all three, even though only the mirage-crypto-pk's code is affected.

I've added mirage-crypto-pk to ensure the new tests that verify the fix are run in CI. It's a shame opam doesn't display the results of the tests. I've manually verified that the added tests are successful (in test_pk_runner):
```
$ dune runtest
File "tests/dune", line 34, characters 34-45:
34 |  (libraries mirage-crypto-entropy mirage-unix))
                                       ^^^^^^^^^^^
Error: Library "mirage-unix" not found.
Hint: try: dune external-lib-deps --missing @runtest
test_random_runner alias tests/runtest
.........
Ran: 9 tests in: 0.11 seconds.
OK
test_pk_runner alias tests/runtest
.......................................................................................................................................................................................................................................................................................................................................................................
Ran: 359 tests in: 1.13 seconds.
OK
```

The patches can be more easily reviewed by looking at https://github.com/psafont/mirage-crypto/compare/v0.6.2...psafont:rsafix-stockholm

~~This is a draft PR as it need to be tested, there were quite a few changes in the library, in particular the change from build-time to run-time hardware acceleration for xor, hashes and rng. xor and hashes being the most relevant.~~


edwintorok:
I think all CPUs on the HCL currently support AES-NI (see http://hcl.vmd.citrix.com/cpus/?cpusupport__version=17&vendor=2 and https://en.wikipedia.org/wiki/AES_instruction_set#Intel), but it might be useful to test on a CPU that doesn't support it just to check that the CPUID detection code works correctly (we had an issue in the past where XAPI crashed with SIGILL immediately on startup due to this when CPUID detection was done at build time).

lindig (approved):

lindig:
This is a backport to Stockholm. Should the testing mentioned by @edwintorok not be done before we merge this on the trunk branch?

edwintorok:
Does trunk already have the new mirage-crypto, or does it require the rebuild of the toolstack after the previous xs-opam update?

psafont:
The build-time to run-time checks were merged in november to trunk using 4ba1c529b1f7490ce7b5f09db48d1b8a87a8625c

We can test trunk right away with hosts that don't support aes-ni and other hw instructions

psafont:
Current failure is due to the newest version needing dune 2.6.0+ to build.

We've using 2.7 in trunk since october: 251272fd1274d58370e191f944b71cf971ecd57c

psafont:
mirage-crypto depends on dune >= 2.6, but the version used in master branch (2.7.1) depends on result 1.5.

Alcotest 0.8.5 does not work with that version of result, and updating alcotest requires changes in some of the toolstack packages.

I think this is bloating this change a lot. I think it would be preferrable to have a patch for the mirage-crypto* libraries that has the correct logic for private keys instead of forcing all these changes. We can think at a later time whether is worth including a bigger batch of updates for a bigger, incremental upgrade for 8.2 hosts.

lindig:
I agree that if we primarily want to fix the key handling, a dedicated patch is the right solution, especially for a backport.

lippirk (approved):

lindig (approved):
Given the sensitive nature of crypto algorithms I'm not qualified to review this. Everything looks good to me but this isn't saying much. I am in support of the general approach of cherry picking commits from a higher release to fix the issue, as has been done here.

PR #550, opened 2021-02-19
maintenance: pin stockholm branches in xs-extra

lippirk:
We do not need to create a new xs-opam release, this is just helpful for development on stockholm


psafont (approved):

psafont:
the rrdd-plugins directory name has `-lcm`, unlike all the others

lippirk:
> the rrdd-plugins directory name has `-lcm`, unlike all the others

that was a mistake (though I don't think it matters functionally) - I've fixed it now :)

lindig:
The purpose is to fix the versions of components whose master branch we used to use but which are not packages into the xs-opam-repo RPM? So this is only relevant for the CI but not RPM packaging.

lindig (approved):

PR #557, opened 2021-04-15
Fix and update upstream-extra

edwintorok:
Upstream-extra was uninstallable due to missing charInfo_width package. Fix that and also upgrade the various packages that can be upgraded (some would require OCaml 4.11 or 4.12 for further upgrade so I added the highest that'd install on 4.10).


lippirk (approved):

packages/upstream-extra/ocaml-lsp-server.1.4.1/opam
```diff
@@ -1,10 +1,7 @@
-x-commit-hash: "154ccd7142651807d40eb851969f318b873b7fb4"
```
psafont, line 1: The version is not changed, are these changes fixing a problem?

The contents of the file are different from the default repo because I use `opam show --raw pkg.ver` to automate the package updates.

packages/upstream-extra/dot-merlin-reader.4.1/opam
```diff
@@ -0,0 +1,30 @@
+opam-version: "2.0"
```
psafont, line 1: Does this version play well with merlin 3.5?

lindig (approved):

psafont:
This pr seems to have stalled a bit I've prepared a couple of fixups to address the comments I made. It should be easy to verify if they are OK and squash + merge the PR

lindig:
@edwintorok can we squash and merge?

PR #563, opened 2021-05-20
ci: detect unused packages

### LINKED ISSUE #334: Detect unused upstream packages

psafont:
Most of them are dependencies for ocamlformat, but others like conf-perl or ppx_fail are not so clear why they are there, thoughts?

Fixes #334 

- conf-perl is used to build zarith, cannot be removed
- cpuid was used by previous versions of mirage-crypto, it's unused
- ppx_fail: no package depends on it on the default repo, on stockholm ppx_jane depended on it (and the whole JST ecosystem)


tools/find-unused-packages.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+USED=$(opam list -t --required-by xs-toolstack --recursive --short | sort)
+UPSTREAM=$(find "packages/upstream/" -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 -I {} echo "{}" | cut -d "/" -f 3 | cut -d "." -f 1 | sort)
+XS=$(find "packages/xs/" -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 -I {} echo "{}" | cut -d "/" -f 3 | cut -d "." -f 1 | sort)
+diff -u .github/unused-upstream.txt <(comm -23 <(echo "$UPSTREAM") <(echo "$USED"))
+diff -u .github/unused-xs.txt <(comm -23 <(echo "$XS") <(echo "$USED"))
```
lippirk, line 9: It'd be nice to print the diff to fix (set -e does not make this easy, silly bash)

```suggestion
UPSTREAM_DIFF=$(diff -u .github/unused-upstream.txt <(comm -23 <(echo "$UPSTREAM") <(echo "$USED")) || true)
XS_DIFF=$(diff -u .github/unused-xs.txt <(comm -23 <(echo "$XS") <(echo "$USED")) || true)

if [ -n "$XS_DIFF" ] || [ -n "$UPSTREAM_DIFF" ]; then
  echo "$XS_DIFF" 1>&2 ;
  echo "$UPSTREAM_DIFF" 1>&2 ;
  echo "ERROR please fix above diffs" 2>&2 ;
  exit 1
fi
```

psafont: the diff is being printed, that's how I iterated on the changes (and saw that the checks were being done too soon)

example: https://github.com/xapi-project/xs-opam/runs/2631749150?check_suite_focus=true

psafont: who doesn't?

lippirk (approved):

.github/workflows/main.yml
```diff
@@ -46,15 +46,18 @@ jobs:
...
-      - name: Check whether there are packages with more than a version
-        run: tools/opam-single-version-check.bash
-
       - name: Upgrade existing packages
         run: |
           opam update
           opam depext -vv -y xs-toolstack
           opam upgrade
 
+      - name: Check whether there are packages with more than a version
```
lindig, line 55: With more than "one" version, I suppose.

lindig (approved):

PR #566, opened 2021-06-18
CP-37034: add tyre: library for typed regular expressions

edwintorok:
Support both parsing and unparsing.
Can be used as replacement for scanf/printf to perform validation.


lippirk (approved):

psafont:
Do we still want to add this library right now?

lindig (approved):
I like that the dependencies of this library are reasonable.

edwintorok:
Even if we don't use this yet it might be useful to have it, to make it easier to write parsers (e.g. I think we suggested to use `angstrom` or `tyre` for parsing the yum command output)

PR #570, opened 2021-07-15
Use the official cache as a mirror

psafont:
This should minimize disruptions cause by server downtime, which is
happening quite often with inria's gitlab instance.

@mseri do you know of any downsides by doing this? I'm not worried about users not being able to configure their mirrors in this case. (https://github.com/ocaml/opam/issues/4411)

### DIFF
--- a/repo
+++ b/repo
@@ -1,4 +1,4 @@
 opam-version: "2.0"
 browse: "https://github.com/xapi-project/xs-opam/tree/master/"
 upstream: "https://github.com/xapi-project/xs-opam/tree/master/"
-archive-mirrors: "cache"
+archive-mirrors: "https://opam.ocaml.org/cache"


edwintorok (approved):

lippirk (approved):

lindig:
What does this do? Packages are not fetched from the URL of their opam description but from the cache first? And this is safe because of the checksums such that the cache can't be poisoned?

psafont:
> Packages are not fetched from the URL of their opam description but from the cache first? And this is safe because of the checksums such that the cache can't be poisoned?

Both are correct AFAIK. The issue about this setting taking priority over the local cache is not affecting our CI, this can be checked by the logs.

edwintorok:
The cache seems to use a weak checksum (MD5), but the opam file (at least for menhir) has both and md5 and sha512 checksum, which should be safe.

mseri:
There is a cache for all checksums fwiw: Try "https://opam.ocaml.org/cache/sha512/XX/XXXXXX...XX"

PR #571, opened 2021-07-15
maintenance: schedule builds for the yangtze branch

psafont:
The builds are scheduled daily between master and stockholm builds

### DIFF
--- a/.github/workflows/yangtze-lcm.yml
+++ b/.github/workflows/yangtze-lcm.yml
@@ -1,0 +1,20 @@
+name: Build and test (Yangtze, scheduled)
+
+on:
+  schedule:
+    # run daily
+    - cron: '51 2 * * *'
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


lippirk:
Stockholm also has some push/PR rules:

```
  push:
    branches:
      - 'release/stockholm/lcm'
    tags:
      - '*'
  pull_request:
    branches:
      - 'release/stockholm/lcm'
```

maybe those would be nice too

lippirk (approved):

psafont:
those are already define in the respective branch: https://github.com/xapi-project/xs-opam/blob/release/yangtze/lcm/.github/workflows/lcm.yml

Remember that this is only about the scheduled builds, which must be defined in the master branch

edwintorok (approved):
Ah yes, you did say that scheduling needs to be configured on master. Now I see what you meant.

PR #574, opened 2021-07-28
maintenance: fix build in opam 2.1.0

psafont:
opam 2.1.0 uses depexts for package resolution so with broken depexts
packages cannot be installed anymore


lippirk:
Did you get these changes from here? https://github.com/ocaml/opam-repository/blob/master/packages/conf-pkg-config/conf-pkg-config.2/opam

Your new version looks the same, but with some of the fields moved around

lippirk (approved):

psafont:
the changes are from upstream, yes. It's not the contents of the upstream git repo, but the output of `opam info --raw` when using the default repository as the source. Note that the name of the depexts have been changed.

lindig (approved):

PR #583, opened 2021-09-06
fix: follow Stockholm branches after Yangtze split

psafont:
This affects message-switch, xapi-idl and xapi, other repositories were
already branched or are still not branched.

### DIFF
--- a/packages/xs-extra/message-switch-async.1.23.1/opam
+++ b/packages/xs-extra/message-switch-async.1.23.1/opam
@@ -1,5 +1,5 @@
 opam-version: "2.0"
-name: "message-switch-lwt"
+name: "message-switch-async"
 maintainer: "xen-api@lists.xen.org"
 authors: [ "xen-api@lists.xen.org" ]
 homepage: "https://github.com/xapi-project/message-switch"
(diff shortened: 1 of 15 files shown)


psafont:
message-switch fails because of unused bindings, I'll try to reduce their severity, this is fixed in yangtxe by removing the lines:

```
# File "switch/traceext.ml", line 19, characters 0-32:
# 19 | type message_id = string * int64
#      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Error (warning 34): unused type message_id.
# File "switch/traceext.ml", line 34, characters 2-26:
# 34 |   type kind = Message.kind
#        ^^^^^^^^^^^^^^^^^^^^^^^^
# Error (warning 34): unused type kind.
```

edwintorok:
Should we keep maintaining Stockholm LCM separately? Might be useful to put into Stockholm whatever we put into Yangtze once Yangtze is released.

edwintorok (approved):

psafont:
I think it's useful for the time being as we still maintain it

lindig (approved):

PR #586, opened 2021-09-23
Backport fixes to yangtze

psafont:
Backports for
- CA-341597: add conf-libev, Raise the open fd limit in varstored v0.6.1
- CA-358583: Fix PROXY metadata to multiple HTTPS requests on a connection in in xen-api-libs-transitional v2.25.3 (already included in the current archive)
- Fix handling of web-dir parameter in xen-api v1.249.15 (already included in the current archive)

### DIFF
--- a/packages/upstream/conf-libev.4-11/files/build.sh
+++ b/packages/upstream/conf-libev.4-11/files/build.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+ocaml discover.ml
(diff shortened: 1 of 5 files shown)


lindig (approved):

lindig:
The "discover.ml" code is not the most modern-looking code but since this is just run like s script for discovery I see no reason to ask for changes.

lindig:
Is this complete? I only see the changes for lebev discovery but not what is mentioned above:

    CA-341597: add conf-libev, Raise the open fd limit
    CA-358583: Fix PROXY metadata to multiple HTTPS requests on a connection
    Fix handling of web-dir parameter

psafont:
I've just updated the varstored-package to use the lcm branch instead of 0.6.0, this should affect the the tarball generated, but it's good to have.

For the other two fixes no changes were necessary as the repositories already follow the respective lcm branches, I'll clarify this on the PR

PR #596, opened 2022-01-31
maintenance: add ocaml 4.12.1 and 4.13.1

psafont:
These are the 2 latest stable versions, let's see what the tests bring up.


psafont:
The CI passing depends on https://github.com/xapi-project/xs-opam/pull/596 being merged

psafont:
4.13.1 is not feasable right now if we want developer tooling to stay the same (ocamlformat)

lindig (approved):

lindig:
Regarding ocamlformat - I would believe that we can't stem the tide. So this change is going to happen sooner or later. But we might want to disable in the CI that checks for the formatting such that we can move over repos one by one.

psafont:
I think moving to 0.19.0 introduces changes, but not that many (it changes how `::` and multilines are formatted and the indentation of closing parameters) This version is compatible with 4.13 and I'm not opposed to adopting it.

For newer versions there are quite a bit of deprecations, we've not been the only ones complaining about some of the options deprecated. I there's a chance that a profile that fits the project may appear in future versions.

edwintorok (approved):
Thanks, having support for new compilers will be useful so we can start taking advantage of new tooling (memtrace). Also with support for 4.13 we can start running the no-naked-pointer checker to prepare for 5.0

psafont:
Note that it will be tough to upgrade to 4.14 because of ocamlformat: 0.20 is the only that supports it and it will be very disruptive.

lindig:
The main problem I see are backports. If we move to a different format, we will always have diffs against older releases.

robhoes:
Yes, that's annoying. Even between Yangtze and master there are changes that affect almost anything you backport. Changes _seem_ to be arbitrary... bracket at the end of the line versus next line, things like that.

PR #588, opened 2021-11-11
Upstream updates

psafont:
Quite a few updates remaining, opening the PR and leaving as draft to check if the tests pass before adding more.

PR #598, opened 2022-02-17
maintenance: drop previous ocaml versions

psafont:
This means also dropping several dev tools and disallowing several
versions of these packages
This also prepares the Ci for xapi to use 4.13


edwintorok (approved):
LGTM, can we also add `memtrace` to take advantage of the memory profiling in newer OCaml versions?

robhoes (approved):

psafont:
> can we also add memtrace to take advantage of the memory profiling in newer OCaml versions?

Sure, I'd like to do that as part of the update in https://github.com/xapi-project/xs-opam/pull/588

PR #600, opened 2022-02-21
CP-38617: expose xentoollog as a package and use it

psafont:
This package is not to be installed in the product and is only used to
ease local development.

Depends on https://github.com/xapi-project/xenctrl/pull/4 being merged

### DIFF
--- a/packages/xs-extra/xentoollog.master/opam
+++ b/packages/xs-extra/xentoollog.master/opam
@@ -1,5 +1,5 @@
 opam-version: "2.0"
-synopsis: "Mock OCaml bindings for the Xen Hypervisor"
+synopsis: "Mock OCaml bindings for Xen's xentoollog"
 maintainer: "xen-api@lists.xen.org"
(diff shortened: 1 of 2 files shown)


edwintorok (approved):

packages/xs-extra/xentoollog.master/opam
```diff
@@ -0,0 +1,18 @@
+opam-version: "2.0"
+synopsis: "Mock OCaml bindings for Xen's xentoollog"
+maintainer: "lindig@gmail.com"
+authors: "lindig@gmail.com"
```
lindig, line 4: Is this correct? Are these the mock bindings? And maybe I should not be named as an author with my private address if this is in `xapi-project`.

psafont: These are the mock bindings, I'll change the maintainer and authors

lindig (approved):

PR #610, opened 2022-09-12
CP-40716: General update of dependencies

psafont:
This needs small adaptations in xapi, see https://github.com/xapi-project/xen-api/pull/4786

Depends on a new version of nbd containing https://github.com/xapi-project/nbd/pull/163

Successful run with the xapi branch pinned: https://github.com/xapi-project/xs-opam/actions/runs/3152874137/jobs/5128646432

Full CH build with the code done, testing it internally in Suite Run ~~175138~~ ~~175209~~ 175221


psafont:
note that there's a commit that pins xapi to the compatible version, which is why the checks are successful. I'll move it to the end of the series, link the successful workflow here as proof then remove the commit to ready it for merging

packages/xs-extra/forkexec.master/opam
```diff
@@ -24,5 +24,5 @@ synopsis: "Sub-process control service for xapi"
 description:
   "This daemon creates and manages sub-processes on behalf of xapi."
 url {
-  src: "https://github.com/xapi-project/xen-api/archive/master.tar.gz"
+  src: "https://github.com/psafont/xen-api/archive/private/paus/refresh.tar.gz"
```
edwintorok, line 27: We'll need to remember to remove this pinning patch those branches are merged

edwintorok (commented):
There are a lot of upstream changes here that will need testing, but once an internal build and test passes I think we can merge this. Just lets remember to undo the pinning afterwards.

edwintorok (approved):

robhoes (approved):

PR #615, opened 2022-10-28
CA-371780: update xapi-rrd to 1.9.1

psafont:
There was a piece of code with quadratic cost that could dominate the runtime in xcp-rrdd in busy hosts

### DIFF
--- a/packages/xs/xapi-rrd.1.9.1/opam
+++ b/packages/xs/xapi-rrd.1.9.1/opam
@@ -1,19 +1,19 @@
 opam-version: "2.0"
 name: "xapi-rrd"
-version: "1.9.0"
+version: "1.9.1"
 synopsis: "RRD library for use with xapi"
 description: """\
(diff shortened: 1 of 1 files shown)


psafont:
I'll cut an xs-opam release after merging this, we still haven't included 1.9.0 into the product, and this is sorely needed

lindig (approved):

edwintorok (approved):

edwintorok:
If we intend to make further optimizations on the Rrd data structures shouldn't we merge this as a subdir into XAPI itself so it is easier to make changes?

psafont:
> If we intend to make further optimizations on the Rrd data structures shouldn't we merge this as a subdir into XAPI itself so it is easier to make changes?

We've had request about keeping the library up-to-date in the upstream opam repository, not sure if there are any actual users or it was somebody curious about the state of the library

PR #619, opened 2022-12-07
ocaml: add 4.14

psafont:
Update packages to be compatible with ocaml 4.14.

Some packages around ocamlformat need to have 2 versions to be able to work in both versions of the compilers, this affects the artifacts that generated from this version.

Because the ocamlformat version has been changed once we move to ocaml 4.14 the formatting of xapi will need to change accordingly

I would rather wait until 4.14.1 is available before merging this, it is probably a good idea to merge the upstream changes before this one as well.

### DIFF
--- a/packages/upstream-extra/merlin.4.6-414/opam
+++ b/packages/upstream-extra/merlin.4.6-414/opam
@@ -1,0 +1,78 @@
+opam-version: "2.0"
+name: "merlin"
+version: "4.6-414"
+synopsis:
+  "Editor helper, provides completion, typing and source browsing in Vim and Emacs"
+description:
+  "Merlin is an assistant for editing OCaml code. It aims to provide the features available in modern IDEs: error reporting, auto completion, source browsing and much more."
+maintainer: "defree@gmail.com"
+authors: "The Merlin team"
+license: "MIT"
+homepage: "https://github.com/ocaml/merlin"
+bug-reports: "https://github.com/ocaml/merlin/issues"
+depends: [
+  "ocaml" {>= "4.14" & < "4.15"}
+  "dune" {>= "2.9.0"}
+  "merlin-lib" {= version}
+  "dot-merlin-reader" {>= "4.6"}
+  "yojson" {>= "2.0.0"}
+  "conf-jq" {with-test}
+]
+conflicts: [
+  "seq" {!= "base"}
+  "base-effects"
+]
+build: [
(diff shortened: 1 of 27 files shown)


tools/no-duplicates-check-upstream-extra-dummy.bash
```diff
@@ -13,7 +13,7 @@ for pkg in $packages; do
   #   ocaml, ocaml-system, ocaml-base-compiler, ocaml-config
   # several versions of the dev tools might be needed:
   #   merlin, ocaml-lsp-server
-  if [ "$versions" -gt 1 ] && echo "$pkg" | grep -vqE '^xenctrl$' ; then
+  if [ "$versions" -gt 1 ] && echo "$pkg" | grep -vqE '^xenctrl$|^ocaml$|^ocaml-base-compiler$|^ocaml-system$|^merlin$|^ocaml-lsp-server$|^ocamlformat$|^ocamlformat-rpc-lib$|^odoc-parser$' ; then
```
lindig, line 16: These patterns could be written as `^(A|B|C)$` or using the `-x` option.

lindig (approved):

edwintorok:
Dune 3.7 is nearly ready, we may want to include that to also get the bugfix for binary corruption: https://github.com/ocaml/opam-repository/pull/23349

packages/upstream-extra/merlin.4.6-414/opam
```diff
@@ -0,0 +1,78 @@
+opam-version: "2.0"
+name: "merlin"
```
edwintorok, line 2: I think merlin is vendored inside ocaml-lsp-server already (with some bugfixes on top), is it useful to keep this separate package, do we still use it directly?

edwintorok (approved):
The changes here look good. We might need to adjust a few of the [tools/scripts ](https://github.com/xapi-project/xs-opam/pull/627) to work with 2 compilers, I can attempt to do those tweaks once the other PR is merged.

packages/upstream-extra/merlin.4.6-414/opam
psafont: I was keeping it in case developers used it, if @lindig is okay with it, it can be dropped

lindig: I am using merlin but for me it does not have to be in XS-opam

PR #620, opened 2022-12-13
xs-opam update to add goblint (needed for static analysis of C stubs)

edwintorok:
Depends on https://github.com/xapi-project/xs-opam/pull/627

This adds quite a lot of packages, mostly to upstream-extra. Although we do gain a static analyzer that can avoid bugs in our C stubs in exchange.


tools/cleanup.sh
```diff
@@ -1,5 +1,16 @@
...
+# so find out recursive deps of xs-toolstack: we want to run tests for all of
+# those (but we don't necessarily need to run the tests of packages pulled in
+# only for testing - that is sometimes not all solvable with same versions)
+set -euo pipefail
+COMPILER=4.14.0
+opam admin check --ignore-test-doc
+# vhd-format-lwt would need io-page 2.4.0 which brings in a lot of other
+# updates, which is being worked on separately
+ALLDEPS=$(opam admin filter --verbose --dry-run --resolve=xs-toolstack | sed -n 2p | tr ' ' '\n' | grep -v vhd-format-lwt | tr '\n' ',')
+exec opam admin filter --dry-run "--resolve=${ALLDEPS}" --resolve upstream-extra-dummy --resolve xenctrl.dummy --resolve opam-depext --resolve opam-ed.0.3 --resolve ocaml-base-compiler.${COMPILER} --resolve ocaml-system.${COMPILER} --or --with-test
```
lindig, line 16: The `--resolve` can take a comma-separated list.

```
RESOLVE += upstream-extra-dummy
RESOLVE += xenctrl.dummy
..
..--resolve=$(echo "$RESOLVE" | tr ' ' ',').
```

The above could work - not tested it.

tools/cleanup.sh
```diff
@@ -1,5 +1,14 @@
...
-#
-exec opam admin filter --dry-run --resolve=xs-toolstack --resolve upstream-extra-dummy --resolve depext.transition --resolve opam-ed.0.3 --resolve ocaml-base-compiler.4.08.1 --resolve ppx_tools.6.0+4.08.0 --resolve ocaml-system.4.08.1 --or --with-test
+
+# with-test only looks at test dependencies of the packages listed on the
+# cmdline, not everything recursively
+# so find out recursive deps of xs-toolstack: we want to run tests for all of
+# those (but we don't necessarily need to run the tests of packages pulled in
+# only for testing - that is sometimes not all solvable with same versions)
+set -euo pipefail
+COMPILER=4.14.0
```
psafont, line 11: This version is not yet available, it would be better not to hardcode it. this value should be taken from the CI variable that determines the ocaml version used, or load the env variables from https://github.com/xapi-project/xs-opam/blob/master/tools/xs-opam-ci.env

tools/installtest.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+#
+# list packages that are not required by packages in this repo
+
+# with-test only looks at test dependencies of the packages listed on the
+# cmdline, not everything recursively
+# so find out recursive deps of xs-toolstack: we want to run tests for all of
```
psafont, line 7: We might want to exclude some packages from being tested, mainly binaries used for development (ocamlformat, merlin, utop and co)

packages/upstream/x509.0.11.2/opam
```diff
@@ -33,7 +33,7 @@ depends: [
   "alcotest" {with-test}
   "cstruct-unix" {with-test & >= "3.0.0"}
   "mirage-crypto-rng" {with-test}
-  "mirage-crypto-pk" {with-test & < "0.8.9"}
+  "mirage-crypto-pk" {with-test}
```
psafont, line 36: The version check is there because of an RSA issue that was fixed in 0.8.9, I'm not sure we can relax this

lindig:
This could use a set of fresh eyes or we should move this into a draft if it can't be progressed.

psafont:
I believe the PR needs to address the issues I pointed out

edwintorok:
Lets wait until the dependencies are updated before merging this, might be difficult to get it to work correctly and pass with the current versions.

edwintorok:
Rebased on top of [the consistent xs-opam PR ](https://github.com/xapi-project/xs-opam/pull/627)

tools/cleanup.sh
edwintorok: Will address in [the other PR ](https://github.com/xapi-project/xs-opam/pull/627)

tools/cleanup.sh
edwintorok: addressed in the other PR: https://github.com/xapi-project/xs-opam/pull/627

packages/upstream/x509.0.11.2/opam
edwintorok: removed this

edwintorok:
A koji chain-build has succeeded.

edwintorok:
Converting back to a draft, if I actually try to use 'goblint' in a .spec file it doesn't work because of a missing dependency in xs-toolstack.

edwintorok:
I think I'll need to move all these packages from upstream-extra to upstream, otherwise they won't be available in xs-opam-repo for xapi

psafont:
Failure is due to a new license being added, we should add instructions on the error in retrospect. It needs to be added to .github/known-licenses.txt

psafont:
This should be ready to rebase, now that the 4.14 changes are in master

psafont (approved):

edwintorok:
This is now ready to be merged

lindig:
@psafont is not here but I suggest to merge it. For now it will only affect CI and we could revert this if anything goes wrong.

PR #625, opened 2023-01-12
ci: avoid deprecated actions in github

psafont:
In particular the "official" actions to set up releases were abandoned, so the gh CLI is used instead

Signed-off-by: Pau Ruiz Safont <pau.ruizsafont@cloud.com>


jameshensmancitrix (approved):

edwintorok (approved):
I think the only way to test this is to run it?

PR #626, opened 2023-01-31
upstream: Add opentelemetry libraries for testing, includes ocurl and ezcurl

psafont:

### DIFF
--- a/packages/upstream/conf-libcurl.2/opam
+++ b/packages/upstream/conf-libcurl.2/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+name: "conf-libcurl"
+version: "2"
+synopsis: "Virtual package relying on a libcurl system installation"
+description:
+  "This package can only install if the libcurl is installed on the system."
+maintainer: "blue-prawn"
+authors: "Daniel Stenberg"
+license: "BSD-like"
+homepage: "http://curl.haxx.se/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+flags: conf
+build: ["curl-config" "--libs"]
+depexts: [
+  ["libcurl4-gnutls-dev"] {os-family = "debian"}
+  ["libcurl-devel"] {os-distribution = "mageia"}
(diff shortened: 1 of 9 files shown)


edwintorok (approved):

lindig:
We might want to add the patch to ezcurl that Edwin proposed upstream (because we need it for vtpm) until it is accepted and released. It's two lines.

lindig (approved):

lindig:
```
commit a92d7bbe4cd0a528ce6710f932d115d0c3f06d73
Author: Edwin Török <edwin.torok@cloud.com>
Date:   Tue Jan 31 16:35:46 2023 +0000

    ezcurl: allow POST with non-form data
    
    Currently ezcurl always sets CURLOPT_HTTPPOST, but that means it is not able
    to send a custom body that is not a form (e.g. JSON is quite a common use-case
    these days).
    
    Add a special case: if the POST params is empty, *and* a 'content' is set, then
    just set CURLOPT_POST, but not CURLOPT_HTTPPOST.
    
    Example usage:
    ```
    let contents = Ezjsonm.to_string json in
    let config = EZ.Config.(verbose true default) in
    match EZ.post ~config ~url ~params:[] ~content:(`String contents) () with
    ```
    
    Signed-off-by: Edwin Török <edwin.torok@cloud.com>

diff --git a/src/core/Ezcurl_core.ml b/src/core/Ezcurl_core.ml
index 09999c3..f26216d 100644
--- a/src/core/Ezcurl_core.ml
+++ b/src/core/Ezcurl_core.ml
@@ -317,6 +317,8 @@ module Make(IO : IO)
     let resp_headers_done = ref false in (* once we get "\r\n" header line *)
     Curl.set_url self url;
     begin match meth with
+      | POST [] when (content <> None) ->
+        Curl.set_post self true
       | POST l -> Curl.set_httppost self l;
       | GET -> Curl.set_httpget self true;
       | PUT -> Curl.set_put self true;
```

PR #629, opened 2023-02-22
Yangtze: update metadata from the official repository and fix license metadata

psafont:
The licenses metadata that was missing or was incorrect was was copied over from the master branch, and in the case where that was also incorrect, I verified the source repository for the license.

I've added an additional check to the licenses to ensure all the licenses are known to be correct and will port this to master as well to make sure we produce consistent licenses there as well.


psafont:
The first commit is autogenerated, the rest were fixes by hand

edwintorok (approved):
LGTM. If there are missing licenses on upstream packages it might be useful to submit a PR to the upstream repo/repositories too so we don't have to do this again the next time we update the packages.

psafont:
For the master branch the upstream repository contains all the licenses, they are not regularised yet, so some inconsistencies are expected there

PR #630, opened 2023-02-22
upstream: enforce only known licenses

psafont:
This helps find licenses that are invalid SPDX identifiers and change them for valid ones.

A list of valid identifiers can be found in https://spdx.org/licenses/


edwintorok:
https://github.com/spdx/license-list-data/blob/main/json/licenses.json there is a json file here that we could extract all the identifiers from

psafont:
The current matching approach is a bit limited, we don't have a way to parse compound licenses created using`WITH` and `AND`.
The current json file with licenses doesn't define these either so we'd be limited in that front if we want to build a parser programmatically

tools/license-check.sh
```diff
@@ -12,6 +12,11 @@ do
     echo "ERROR: Package $name does not have a license"
     errors=true
   fi
+  license=$(grep -Po 'license: \K.*' $file| tr -d '"')
+  if echo $license | grep -vqEx 'Apache-1.0|Apache-2.0|BSD-2-Clause|BSD-3-Clause|curl|GPL-1.0-or-later|GPL-2.0-only|GPL-3.0-only|ISC|MIT|MPL-2.0|PSF-2.0|LGPL-2.0-or-later|LGPL-2.1-only|LGPL-2.0-only WITH OCaml-LGPL-linking-exception|LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception|LGPL-2.1-only WITH OCaml-LGPL-linking-exception|LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception|LGPL-3.0-only WITH OCaml-LGPL-linking-exception|LGPL with OpenSSL linking exception|GPL-2.0-only AND LGPL-3.0-only|MIT AND LGPL-2.1-only WITH OCaml-LGPL-linking-exception'; then
```
lindig, line 16: I would prefer the licenses in a list but I don't have a good idea to do this in a shell script. One way would be to store them in a file and use `grep -f`

lindig (approved):

tools/license-check.sh
psafont: I've put the licenses into file, hopefully this is more palatable now

.github/known-licenses.txt
```diff
@@ -0,0 +1,22 @@
...
+MPL-2.0
+PSF-2.0
+LGPL-2.0-or-later
+LGPL-2.1-only
+LGPL-2.0-only WITH OCaml-LGPL-linking-exception
+LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception
+LGPL-2.1-only WITH OCaml-LGPL-linking-exception
+LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception
+LGPL-3.0-only WITH OCaml-LGPL-linking-exception
+LGPL with OpenSSL linking exception
```
edwintorok, line 20: We've looked and this is LGPL-2.1, for now lets keep it like this if it matches upstream

edwintorok (approved):

.github/known-licenses.txt
```diff
@@ -0,0 +1,22 @@
+Apache-1.0
+Apache-2.0
+BSD-2-Clause
+BSD-3-Clause
+curl
+GPL-1.0-or-later
+GPL-2.0-only
+GPL-3.0-only
```
edwintorok, line 8: This would be quite worrying, but it looks like it is only conf-autoconf, only used during build and not at runtime.
In general having both GPL-3.0-only and GPL-2.0-only would be a problem if we ever linked them into the same binary...

PR #643, opened 2023-06-05
ocaml: harden against patchset checksum changing

psafont:
github did the old switcheroo and changes the context of a patch file. This is probably because github doesn't use full commit indeces by default in these file and it needed longer hashes to differenciate between commits.
This change is needed to pass the CI, it doesn't affect the tarball that's being packaged for releases


edwintorok:
The failure seems to be for the sha256 of a patch though, not the tarball.

edwintorok:
Can you add all the .patch files into this repo so that we wouldn't fail fetching them if github changes them in subtle ways?

edwintorok:
commit 5095d4fd3728a0b7783c24f4b45dbe663c9f4d7b
Author: David Allsopp <david.allsopp@metastack.com>
Date:   Wed May 31 17:36:21 2023 +0100

    Use ?full_index=1 when referencing GitHub commits

    Undocumented switch causes patches produced to use git-show(1)'s
    `--full-index` option, which guarantees the stability of the patch files
    from GitHub.

jameshensmancitrix (approved):

PR #655, opened 2023-08-21
Update upstream packages

psafont:
None of the changes should break the current toolstack as they are minor changes.

Breaking updates that were excluded from this PR:
- mirage-crypto: RNG initialization needs now an explicit PRNG module
- ocamlformat formats code in a different way

Follows https://github.com/xapi-project/xs-opam/pull/654

Draft to see changes inquality gates and fix them


psafont:
Current state of held-off packages and missing updates:

```
$ opam repo add default; and opam repo priority default 1; env OPAMNO=true opam upgrade -a ssl.0.6.0 base.v0.15.1 tar.2.3.0 ocurl.0.9.1 mirage-crypto.0.10.7 ocamlformat.0.22.4 | grep -E "↗|∗"
  ↗ ocamlformat-rpc-lib        0.22.4 to 0.26.0
Proceed with ↻ 88 recompilations and ↗ 1 upgrade? [y/n] n

```

edwintorok:
It is probably best to keep ocamlformat and ocamlformat-rpc-lib versions the same, what prevents that ocamlformat-rpc-lib upgrade from working?

packages/xs/polly.0.4.1/opam
```diff
@@ -0,0 +1,31 @@
+opam-version: "2.0"
+name: "polly"
+version: "0.4.1"
```
edwintorok, line 3: There were quite a few changes to the C bindings here, I think there might still be some outstanding PRs on the upstream repo that I didn't review, I'll try to take another look.

packages/upstream/tar.2.3.0/opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "tar"
-version: "2.0.1"
+version: "2.3.0"
```
edwintorok, line 3: Would be useful to add a note on how the new tar 2.5.1 packages are buggy (e.g. links to issues/PRs) so we can check whether they are fixed in the future.

psafont: Hmm, I think the problematic version was 2.5.0, and the fix was released in 2.5.1

psafont:
> It is probably best to keep ocamlformat and ocamlformat-rpc-lib versions the same, what prevents that ocamlformat-rpc-lib upgrade from working?

I probably works, but I'd rather keep them at the same version, sometimes the version bounds in the dependent packages can be too lax. SInce it's not a critical revision of the library, I'd rather hold it off from upgrading

lindig (approved):

psafont:
Passed a toolstack BVT + BST test, this needs another review

snwoods (approved):

PR #659, opened 2023-10-16
Upgrade stdext to 4.11.3

lindig:
We update stdext for CA-382014 to 4.11.3; this needs to be picked up by xs-opam. Several packages are using the same tar ball and so we have changes in various package definitions.


psafont (approved):
All 11 packages in stdext have been updated, maybe squash the commits?

robhoes (approved):

PR #667, opened 2024-02-28
Add patch for XSI-1457 to ocaml-vhd

lindig:
We are patching an upstream project by adding a patch to the opam file. The original is here:

https://github.com/xapi-project/xen-api/commit/db4d465f16745892b4d4ff37d44f8dbd72182314

### DIFF
--- a/packages/upstream/vhd-format-lwt.0.12.1/files/0001-XSI-1457-Limit-number-of-sectors-to-coalesce.patch
+++ b/packages/upstream/vhd-format-lwt.0.12.1/files/0001-XSI-1457-Limit-number-of-sectors-to-coalesce.patch
@@ -1,0 +1,99 @@
+From e2de1b257ecad00173672411870695f7b362b2fe Mon Sep 17 00:00:00 2001
+From: Christian Lindig <christian.lindig@cloud.com>
+Date: Tue, 27 Feb 2024 15:19:07 +0000
+Subject: [PATCH 1/1] XSI-1457: Limit number of sectors to coalesce
+
+We limit the number of sectors to coalesce when generating a list of
+elements to be copied. The limit is currently set to 4 MiB. This allows
+the progress bar to be displayed at a finer granularity rather than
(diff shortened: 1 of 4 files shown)


packages/upstream/vhd-format-lwt.0.12.0/files/0001-XSI-1457-Limit-number-of-sectors-to-coalesce.patch
```diff
@@ -0,0 +1,99 @@
...
++     been coalesced so far. It is made optional because we only use it when we
++     continuously match with one pattern, i.e. the pattern where we coalesce
++     consecutive sectors, and default it to 1 when we are not coalescing. *)
++  let rec coalesce_request ?(coalesced_sectors = 1L) acc s =
+     let open Int64 in
+-    s >>= fun next -> match next, acc with
++    s >>= fun next ->
++    match (next, acc) with
+     | End, None -> return End
+-    | End, Some x -> return (Cons(x, fun () -> return End))
```
Vincent-lau, line 53: why is it deleting this line and then add the same one back in line 67?

Vincent-lau (approved):

packages/upstream/vhd-format-lwt.0.12.0/files/0001-XSI-1457-Limit-number-of-sectors-to-coalesce.patch
lindig: I would not focus on the shape of the diff and more on the result. I had to do this manually and first re-formatted the existing code and then changed it. So it's important not to miss something.

psafont:
I could have sworn I had added a comment already. Please change the version to 0.12.0+1 to make it obvious there's a patch applied when looking at the version

lindig:
I've bumped the version to 0.12.1 @psafont

psafont (approved):

PR #669, opened 2024-03-01
Update upstream dependencies

psafont:
Wide update of dependencies, All of them seem minor, but needs testing.

I'm holding off known breaking updates for:
- ssl: newer versions require openssl 1.1.0
- JST's base: holding off updated because I want to delete most of them first
- ocurl: currently released version is known to be buggy
- mirage-crypto: rng initialization changed the interface, needs code changes 
- qcheck-core: newer version is incompatible with the deprecated qcheck, which is needed for testing packages


lindig:
A failure in the CI about an unknown dependency `goblint`

psafont:
I think when goblint was included conf-ruby was deleted from the opam metadatada.

I'm trying to maintian that dependency by adding conf-ruby to `upstream-extra` so the package is available in dev environment, but not required to build the tarball.

psafont:
I've made builds, and now I'm testing them. I'm wary about the tar package, so I've started the storage tests, which have more import and export tests:
suite runs 195246, 195249, 195250

lindig (approved):

psafont:
There were two unrelated storage issues, due to lab issues, and another one because the testing code was not up-to-date with the current code in all master branches.

This can be merged whenever we want

Vincent-lau (approved):

PR #671, opened 2024-04-10
upstream-extra: update dev tools

psafont:
No changes in the distribution tarball


lindig (approved):
All in extra, which we don't package.

psafont:
I've added a fix for the broken CI, now all the xapi-based metadata is up-to-date

Vincent-lau (approved):

PR #675, opened 2024-05-10
upstream: non-breaking dependencies

psafont:
Most of them have small fixes, or add ocaml 5.2 compatibility. I want to keep changes to a minimum before using ocaml 4.14.2, I'll force the new interface for mirage-crypto on the next update

### DIFF
--- a/packages/upstream-extra/ocamlformat-rpc-lib.0.26.2/opam
+++ b/packages/upstream-extra/ocamlformat-rpc-lib.0.26.2/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "ocamlformat-rpc-lib"
-version: "0.26.1"
+version: "0.26.2"
 synopsis: "Auto-formatter for OCaml code (RPC mode)"
 description:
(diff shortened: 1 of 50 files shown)


packages/upstream-extra/ocamlformat-rpc-lib.0.26.2/opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "ocamlformat-rpc-lib"
-version: "0.26.1"
+version: "0.26.2"
```
edwintorok, line 3: Do we need to update .ocamlformat anywhere?

edwintorok (approved):

packages/upstream-extra/ocamlformat-rpc-lib.0.26.2/opam
psafont: No, because it introduced a lot of friction at the time. There are not formatting changes between the versions

robhoes (approved):

PR #681, opened 2024-06-13
Metadata refresh, some minor updates, removal of ocaml 4.13.1

psafont:
The metadata updates need to be tested before merging, they changed where the files are hosted. Maybe we can do the same for our patches? It will deduplicate the patches we have for rpclib

As always, it's better to review the commits one by one.


psafont:
The current failure is because of a test in xapi not being attached to any package, fixed by: https://github.com/xapi-project/xen-api/pull/5684

lindig (approved):

packages/ocaml/ocaml-base-compiler.4.13.1/opam
```diff
@@ -1,43 +1,94 @@
 opam-version: "2.0"
```
edwintorok: Do we still need 4.13.1 here? Is it for LCM?

packages/ocaml/ocaml-base-compiler.4.14.1/opam
```diff
@@ -2,7 +2,10 @@ opam-version: "2.0"
 name: "ocaml-base-compiler"
```
edwintorok, line 2: Aren't we on 4.14.2? Should we drop 4.14.1?

edwintorok (approved):
Minor comment about dropping 4.14.1

packages/ocaml/ocaml-base-compiler.4.14.1/opam
psafont: Yes, eventually, some people might not have updated their opam switches to ocaml 4.14.2, which means bad things will happen when the version is removed.

psafont:
This is being tested in suite run 201807

PR #682, opened 2024-06-21
IH-583: Add new dependencies of the standalone systemd implementation

last-genius:
Depends on https://github.com/xapi-project/xen-api/pull/5721/

Added `unix-errno` and `astring` as dependencies for the standalone implementation - these are already packaged.


psafont:
In order to have a smooth transition, it's better to split this merge into two: one that adds the required dependencies to unix-errno and astring, and another one with the removal of the systemd library. The latter can be opened and merged after the xen-api PR is merged

last-genius:
Split this PR in two. The one to remove systemd is #684

psafont (approved):

PR #689, opened 2024-07-05
tools: clean up directory

psafont:
It had quite a few unmaintained files


robhoes (approved):

edwintorok (commented):
Some of the scripts look useful, are they broken, or is it just that we haven't used them in a while?

psafont:
We haven't used them in a while, and they don't add value to users, or maintainers:
- Dockerfile: it's been years since anybody tried to use this, the ocaml dockerfiles ecosystem as moved forwards, so I doubt it works anymore
- The sync from upstream script is unused at the time being, it will break when the default repository structure is adopted.
- The cleanup script will be superceded with the changes to use package metadata dependencies to generate the tarball, and the CI already takes care of unused packages.
- The install test is very brittle, and for little benefit. Brittle because not all test-dependencies might be conisntallable, and little benefit because upsream opam already tests all the packages in there.

PR #693, opened 2024-07-23
Patch Uri packages to add path_unencoded

last-genius:
Until Uri is versioned out into a new tag, we will need to use this patched version pointing to a commit hash. This is required for the xen-api percent-decoding PR.


psafont (approved):
The hash matches the merge commit for the patch that includes the unencoded function: https://github.com/mirage/ocaml-uri/commit/8d6da91636c3d81ecc126552664658bc715b4f14

lindig (approved):
Didn't know it was possible to point to a commit for an archive. This makes it a lot easier than using patches.

PR #696, opened 2024-09-13
CA-399172: fix potential crash in Uri.of_string

edwintorok:
Using Lazy from a multithreaded program should raise Lazy.Undefined. However due to bugs in the OCaml runtime this was actually crashing.

Update to angstrom 0.16.1 which no longer uses Lazy and avoids this crash.

This was initially discovered:
* https://github.com/mirage/ocaml-uri/issues/178

Fixed in Angstrom:
* https://github.com/inhabitedtype/angstrom/pull/229

Runtime bug reported with potential fix from maintainers:
* https://github.com/ocaml/ocaml/issues/13430
* https://github.com/ocaml/ocaml/pull/13434

There are various other usages of Lazy, but I couldn't get those to crash yet, so lets fix the known crash for now, and audit/fix the rest next.


packages/upstream/angstrom.0.16.1/opam
```diff
@@ -1,20 +1,15 @@
...
-exposes monadic and applicative interfaces for composition, and supports
-incremental input through buffered and unbuffered interfaces. Both interfaces
-give the user total control over the blocking behavior of their application,
-with the unbuffered interface enabling zero-copy IO. Parsers are backtracking by
-default and support unbounded lookahead."""
 maintainer: "Spiros Eliopoulos <spiros@inhabitedtype.com>"
-authors: "Spiros Eliopoulos <spiros@inhabitedtype.com>"
-license: "BSD-3-Clause"
+authors: [ "Spiros Eliopoulos <spiros@inhabitedtype.com>" ]
+license: "BSD-3-clause"
```
psafont, line 4: The SPDX identifier shouldn't be changed, it will make the CI to fail. The correct identifier was being used previously: https://spdx.org/licenses/BSD-3-Clause.html

lindig (approved):

packages/upstream/angstrom.0.16.1/opam
edwintorok: Although https://spdx.org/licenses/BSD-3-clause.html redirects to https://spdx.org/licenses/BSD-3-Clause.html, so I think both might be valid.

psafont:
Why was this merged? CI was failing :(

edwintorok:
Hmm I didn't click the button, but github says I did

packages/upstream/angstrom.0.16.1/opam
edwintorok: This says https://github.com/david-a-wheeler/spdx-tutorial they are case insensitive.

edwintorok:
The CI seems to be passing now though 
Screenshot 2024-09-13 at 14 35 31

packages/upstream/angstrom.0.16.1/opam
edwintorok: Opened PR here to see if we can match these in case insensitive way https://github.com/xapi-project/xs-opam/pull/697

psafont: I'll have to change the CI to match the case insensitiveness.

It's tricky to add this changer without adding churn to the licenses, both here and the .spec file (because git is case-sensitive)

PR #697, opened 2024-09-13
[maintenance]: SPDX license identifiers are not case sensitive

edwintorok:
https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/#d2-case-sensitivity says:
> License expression operators (AND, OR and WITH) should be matched in a case-sensitive manner.
> License identifiers (including license exception identifiers) used in SPDX documents or source code files should be matched in a case-insensitive manner

### DIFF
--- a/tools/license-check.sh
+++ b/tools/license-check.sh
@@ -13,7 +13,7 @@ do
     errors=true
   fi
   license=$(grep -Po 'license: \K.*' $file| tr -d '"')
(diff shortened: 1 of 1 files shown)


psafont (approved):
This will need some magic in the xs-opam-repo.spec to not duplicate the licenses

lindig (approved):

PR #703, opened 2024-10-01
xs-extra: update metadata from xen-api.git

psafont:
This updates the metadata to be up-to-date with xapi's


last-genius (approved):

lindig (approved):

psafont:
xapi-tools fails to build because of
```
File "ocaml/xenopsd/scripts/dune", line 5, characters 3-6:
5 | 		(vif as vif)
       ^^^
```

Not sure why, the file is present in that directory

psafont:
The failure is very annoying, I can reproduce it when executing the installation of the package with opam, but now when executing the dune command that opam supposedly executes. This seems to be a problem with the environment of some sort that fails to find the vif file

PR #704, opened 2024-10-02
Change repository to use the default, upstream file layout

psafont:
This needed changes in the tooling, as well as the newest version of opam to produce correct cache and tarballs

The change allows the tarball packaging to be directed purely by metadata, avoiding unused and duplicated packages.

This also allows users to contribute to this opam repository in a similar way to how they do to upstream repository

Updates to the repository in this new form can be seen at this branch: https://github.com/psafont/xs-opam/compare/reorg...psafont:xs-opam:plusplus


robhoes (approved):
It seems good to use the standard layout. Besides that, if it works, it works :)

lindig (approved):

lindig (approved):

PR #707, opened 2024-10-08
Add patch for python cmdline parser in rpclib

Vincent-lau:


Vincent-lau:
Hold on merging as I make a build and test this

psafont:
Do we want to start storing the patches in another repository? It would avoid carrying multiple copies of the patches around.

robhoes:
This appears to be obsolete...

PR #710, opened 2024-10-10
IH-710: update upstream dependencies

psafont:
Only 3 sets of packages are held off:
- ssl because xenserver is not openssl 1.1.1 yet
- mirage cryptography: Changes the hash types to use the ones from a new library, and changes cstructs with string.
- tar: The reader interface was changed in a way that causes non-trivial changes, the writing interface is more manageable. On top of that, it changes cstructs with strings, which can mean it can introduce many unwanted copies. It will require more time and effort until it can be updated, maybe with yet more changes to upstream.


lindig (approved):

psafont:
These packages can build all the toolstack RPMs without any changes.

I'm running the toolstack tests, just in case

robhoes (approved):

psafont:
Tests are all green

PR #718, opened 2025-02-14
Add ocaml-system-4.14.3 to test dev version

edwintorok:
Enable testing the latest OCaml 4.14 branch, which will report itself as 4.14.3.

The opam file in ocaml-system expects an exact match in gen_ocaml_config.ml though, so include the full version in the package version.
OTOH including the full package version in the 'ocaml' package fails with a version  mismatch (that one wants just major.minor.patch).

This is just for testing purposes for now to test the latest OCaml 4.14 branch from github, when you have it installed as a system compiler.


psafont (approved):
It would be interesting if the patches could be added to the PR and enable testing in the CI

lindig (approved):

robhoes:
Is there any reason not to progress this? Are we waiting for the release of 4.14.3?

PR #719, opened 2025-02-19
packages: Update opentelemetry

last-genius:
Removes ambient-context package, adds atomic and thread-local-storage as new dependencies.


last-genius:
`atomic` seems to be a compat package instead of stdlib.Atomic module, which we should have available. Should I remove it entirely, or how does this work?

packages/atomic/atomic.0.1/opam
```diff
@@ -0,0 +1,28 @@
...
+synopsis:
+  "Compatibility package for OCaml's Atomic module, from 4.12"
+author: "OCaml maintainers"
+maintainer: "simon.cruanes.2007@m4x.org"
+license: "LGPL-2.1"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.03" & < "4.12"}
```
psafont, line 13: These versions of ocaml are not available in this repository. Can the dependency from opentelemetry be dropped altogether?

psafont:
THe problem with the license is not only that is not recognized, but it doesn't define. asingle license:
Unrecognised license used for     atomic: LGPL-2.1. Is it a valid a SPDX identifier?

it should be `LGPL-2.1-only` or `LGPL-2.1-or-later`

last-genius:
Sad! Does not seem like we can drop `atomic`. No big deal, will update `opentelemetry` some other time.

edwintorok:
also I think ambient-context will be used by Apollo in a PR fairly soon

PR #727, opened 2025-05-13
github: use newer version of ubuntu

psafont:
The current version is being deprecated

### DIFF
--- a/.github/workflows/lcm.yml
+++ b/.github/workflows/lcm.yml
@@ -11,7 +11,7 @@ on:
       - 'release/yangtze/lcm'
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/lcm.yml
```diff
@@ -11,7 +11,7 @@ on:
       - 'release/yangtze/lcm'
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
edwintorok, line 14: There is also 24.04, shouldn't we switch to the latest?

lindig (approved):

edwintorok (approved):

PR #731, opened 2025-06-03
packages: update mirage-crypto to 2.0.1

psafont:
BREAKING: replaces cstructs for strings, and changes rng generators' API

matching xapi branch is at https://github.com/xapi-project/xen-api/compare/master...psafont:xen-api:private/paus/stringy-crypto


psafont:
> What does it mean to replace ctruct with strings?

In the context of the package it means cstructs are removed from the dependency, and its API now uses strings instead. This does not mean that the cstruct package is removed from the repository because it has a large use-base.

lindig (approved):

PR #732, opened 2025-06-04
Add `qcow-stream` and its dependencies

last-genius:
Adds a library offering QCOW streaming import (that will be used in xapi's vdi-import) and its dependencies.


psafont (approved):
We probably want to change qcow to use [metrics](https://ocaml.org/p/metrics/latest) instead of prometheus

last-genius:
> We probably want to change qcow to use [metrics](https://ocaml.org/p/metrics/latest) instead of prometheus

Indeed, but `prometheus` is also useful for other nefarious purposes :smiling_imp:

lindig (approved):

